PR #3, opened 2017-08-16
Fix build on OSX

benlodotcom:
Thanks for tackling the integration of libvterm with emacs !

This PR should fix building the module on OSX, forkpty is available through util.h on OSX: http://www.unix.com/man-page/osx/3/forkpty/

A neater fix would be to start using autotools to build the module and in the configure phase set some ENV variables depending on the presence of one or the other lib on the OS, but this should do for now.

Test plan:
- Make on OSX => no errors, run vterm-create => terminal shows up
- Make on linux => no errors

### DIFF
--- a/vterm-module.c
+++ b/vterm-module.c
@@ -1,6 +1,10 @@
 #include "vterm-module.h"
 #include <fcntl.h>
-#include <pty.h>
+#ifdef __APPLE__
+#  include <util.h>
+#else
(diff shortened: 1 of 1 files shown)


akermu:
Thanks for the contribution! For now I think a Makefile is enough. If there are more workarounds necessary in the future, I will consider switching to a more sophisticated build system.

PR #21, opened 2018-10-25
trying support scrollback 

jixiuf:
it works now .

and you can use C-v and M-v scroll the screen now without tmux and screen


akermu:
Thanks!

PR #22, opened 2018-10-26
only refresh invalid lines

jixiuf:


VanLaser:
Assuming there are no problems with this code and it gets integrated, this means only the mouse handling remains to be added? These latest emacs-libvterm pulls really seem to be moving the project to a beta/usable stage :) :+1:

vterm.el
```diff
@@ -187,10 +197,22 @@ be send to the terminal."
...
+
+
+(defun vterm--run-timer (buffer)
+  (interactive)
+  (let ((inhibit-redisplay t)
+        (inhibit-read-only t))
+    (when (buffer-live-p buffer)
+      (with-current-buffer buffer
+        (vterm--update vterm--term)))))
+
```
akermu: Is there any particular reason to change the event-driven I/O to timer-based polling?

jixiuf: I think don't need to response to  all I/O  event .

TheBB: My terminals are by and large idle 99% of the time. I think the event-driven updates were fine, and nothing about that design prevents us from ignoring updates on certain events if that is ever a problem.

jixiuf: ok the timer is removed

akermu:
tmp

PR #23, opened 2018-11-08
make other keybinding with prefix like super- hyper- works.

jixiuf:


vterm.el
```diff
@@ -129,13 +129,23 @@ be send to the terminal."
...
 (define-key vterm-mode-map [mouse-3] nil)
 (define-key vterm-mode-map [mouse-4] #'ignore)
 (define-key vterm-mode-map [mouse-5] #'ignore)
-(dolist (prefix '("M-" "C-"))
+(define-key vterm-mode-map [(tab)]  #'vterm--self-insert)
+(define-key vterm-mode-map (kbd "TAB")   #'vterm--self-insert)
+(define-key vterm-mode-map (kbd "<backspace>")   #'vterm--self-insert)
+(define-key vterm-mode-map (kbd "M-<backspace>")   #'vterm--self-insert)
+
+(defvar vterm-esc-map (make-keymap))
```
deb0ch: What is the purpose of making this and binding it to `\e` ?

vterm.el
```diff
@@ -159,6 +169,17 @@ be send to the terminal."
...
         (vterm--update vterm--term key shift meta ctrl)))))
 
+
+(defun vterm-send-key (key &optional shift meta ctrl)
+  "Sends  key to libvterm."
+  (when vterm--term
+    (let* ((inhibit-redisplay t)
+           (inhibit-read-only t)
+           (key key))
+      (when (and shift (not meta) (not cctrl))
```
deb0ch: Is `cctrl` a typo ?

vterm.el
deb0ch: Ok, I think I figured and I think it is to allow passing the `ESC` key to vterm (?). We should be careful that is does not conflict with `evil`, which relies heavily on the `ESC` key to change modes

deb0ch:
On `master` I can navigate `htop` using the arrow keys but not on this PR 😕

vterm.el
jixiuf: > Also, is there a reason why `make-keymap` instead of `make-sparse-keymap`? Conventional Emacs wisdom is to always use `make-sparse-keymap` unless you know for a fact you need `make-keymap`, which is almost never the case.

ok changed to  `make-sparse-keymap`

vterm.el
jixiuf: find

jixiuf:
> On `master` I can navigate `htop` using the arrow keys but not on this PR 😕

fixed

jixiuf:
```
(define-key vterm-mode-map [escape] 'vterm--self-insert)
```
or 
```
(defvar vterm-esc-map (make-sparse-keymap))
(define-key vterm-mode-map "\e" vterm-esc-map)
(define-key vterm-mode-map "\e\e\e" '(lambda()(interactive) (vterm-send-key "\e" )))

```

I don't known which is better

vterm.el
VanLaser: I can't quit `htop` or `less` using ESC even in a normal terminal? (outside emacs).

VanLaser:
@deb0ch  Make sure you test with evil mode set to emacs state (`C-z`) or disabled, so you won't report issues unrelated to vterm.

deb0ch:
Just went back to this branch to test again and now it's working, idk what I missed earlier 😕

I recompiled the C and elisp everytime and restarted Emacs. Strange...

Sorry for the noise.

jixiuf:
```
(defcustom vterm-keymap-exceptions '("C-x" "C-u" "C-g" "C-h" "M-x" "M-o" "C-v" "M-v" "C-c")
  "Exceptions for vterm-keymap.

If you use a keybinding with a prefix-key that prefix-key cannot
be send to the terminal."
  :type '(repeat string)
  :group 'vterm)

(define-key vterm-mode-map (kbd "C-c C-c")   #'(lambda()(interactive) (vterm-send-key "c" nil nil t)))
```
how about add this to vterm.el 
make `C-cC-c` send `C-c`

deb0ch:
Great idea, I did this to integrate it to Spacemacs (still in a PR) and it works like a charm: https://github.com/syl20bnr/spacemacs/pull/11552/files#diff-815f7bec93942d8ed11ccc28e0627130R299

You just need to overwrite `vterm-keymap-exceptions` and map `C-c C-c` to `vterm--self-insert`, no need for a lambda.

There are other things in this PR to integrate `vterm` nicely with other packages like `evil`, `emacs-purpose`, `centered-cursor-mode`, `shell-pop` and others. Maybe it can be some inspiration for stuff to add to this package ? 🤔

vterm.el
deb0ch: Ooooops my bad 😅

deb0ch: I had a conflict with `evil` but it is because `evil` binds ESC for itself.

I feel like evil integration should be the concern of a separate PR.

vterm.el
```diff
@@ -129,13 +129,26 @@ be send to the terminal."
...
+(define-key vterm-mode-map [left]   #'vterm--self-insert)
+(define-key vterm-mode-map [right]   #'vterm--self-insert)
+(define-key vterm-mode-map [up]   #'vterm--self-insert)
+(define-key vterm-mode-map [down]   #'vterm--self-insert)
+(define-key vterm-mode-map [home]   #'vterm--self-insert)
+(define-key vterm-mode-map [end]   #'vterm--self-insert)
+(define-key vterm-mode-map [escape] 'vterm--self-insert)
+(define-key vterm-mode-map [remap self-insert-command] 'vterm--self-insert)
+
+(dolist (prefix '("M-" "C-") )
```
deb0ch: This line should not be modified, but you added an extra space.

vterm.el
```diff
@@ -129,12 +129,25 @@ be send to the terminal."
                           :sentinel #'ignore))))
 
 ;; Keybindings
-(define-key vterm-mode-map [t] #'vterm--self-insert)
+;; (define-key vterm-mode-map [t] #'vterm--self-insert)
 (define-key vterm-mode-map [mouse-1] nil)
```
akermu: Can you remove this line instead of commenting it out?

vterm.el
```diff
@@ -159,6 +172,18 @@ be send to the terminal."
...
+
+(defun vterm-send-key (key &optional shift meta ctrl)
+  "Sends  key to libvterm."
+  (when vterm--term
+    (let* ((inhibit-redisplay t)
+           (inhibit-read-only t)
+           (key key))
+      (when (and shift (not meta) (not ctrl))
+        (setq key (upcase key)))
+      (vterm--update vterm--term key shift meta ctrl))))
```
akermu: This function isn't actually used, is it?

PR #27, opened 2018-11-10
Only refresh invalid lines again

jixiuf:


vterm-module.c
```diff
@@ -138,125 +138,124 @@ static size_t get_col_offset(Term *term, int row, int end_col) {
...
-    }
-    delete_lines(env, liner, 1, false);
-    goto_line(env, liner);
-    refresh_row(term, env, r, width, false);
-  }
+  goto_line(env, line_start);
+  delete_lines(env, line_start, term->invalid_end - term->invalid_start, true);
+  refresh_lines(term, env, term->invalid_start, term->invalid_end, width);
+
+end:
```
akermu: Can you change this to a if statement?

deb0ch:
What does this PR do exactly ?

VanLaser:
Optimization, don't redraw everything, only the modified lines.

deb0ch:
I can confirm that the performance boost is _visible_.

Input used to be very laggy before this PR was merged when the previous commands had a lot of output, now it is blazing fast ⚡️

I just tested again on the commits before and after, the diff is huge.

PR #28, opened 2018-11-10
Use CMake library search to look for libvterm location

knazarov:
On some platforms, the location of libvterm headers and library are
not in standard places that gcc/clang looks for by default.

The usual practice is to use find_path/find_library and then
link them properly. In my particular case, this helps clang to
find homebrew-installed libvterm on OS X.


akermu:
Thanks!

PR #29, opened 2018-11-10
trying support term.el style directory/host/user tracking.

jixiuf:
Notice: for directory/host/user tracking you need to have something
 like this in your shell startup script (this is for a POSIXish shell
 like Bash but should be quite easy to port to other shells)

https://www.emacswiki.org/emacs/AnsiTermHints

for zsh
```
HOSTNAME=$(uname -n)
USER=$(whoami)
case $TERM in
	eterm*)         # emacs
		# The \033 stands for ESC.
		# There is a space between "AnSiT?" and $whatever.
        chpwd() {
            print -P "\033AnSiTc %d"
        } 
        # https://www.emacswiki.org/emacs/AnsiTermHints
        precmd() {
        #     if [[ -n "$SSH_CONNECTION" ]]; then
        #     else
        #     fi
            echo -e "\033AnSiTu" "$USER" # $LOGNAME is more portable than using whoami.
            echo -e "\033AnSiTc" "$PWD"
            echo -e "\033AnSiTh" "$HOSTNAME"
        }
esac
```

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -126,7 +126,7 @@ value with `add-hook'."
   (setq-local scroll-margin 0)
 
   (add-hook 'window-size-change-functions #'vterm--window-size-change t t)
-  (let ((process-environment (append '("TERM=xterm") process-environment)))
+  (let ((process-environment (append '("TERM=eterm-color") process-environment)))
     (setq vterm--process (make-process
                           :name "vterm"
                           :buffer (current-buffer)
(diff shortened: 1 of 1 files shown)


deb0ch:
If I understood correctly, what this is supposed to do is that when you are visiting a remote/local directory in your terminal and from this buffer you use `C-c C-f`, the default location for this `find-file` command will be your shell's current directory ?

Is there anything else that I am missing ?

akermu:
@deb0ch This is correct.

akermu:
@jixiuf Hello, thanks again for the contribution! Is it possible to add the same functionality via a `vterm_parser_callback`? I really don't like to do so much stuff in elisp during the event loop only for directory tracking.

jixiuf:
ok ,I will try  `vterm_parser_callback`

PR #30, opened 2018-11-14
check buffer-live-p in vterm--sentinel

jixiuf:
without this patch I would get 
```
error in process sentinel: save-current-buffer: Selecting deleted buffer
error in process sentinel: Selecting deleted buffer
```
if I set 
```
(defun vmacs-vterm-hook()
  (let ((p (get-buffer-process (current-buffer))))
    (when p
      (set-process-query-on-exit-flag p nil))))

(add-hook 'vterm-mode-hook 'vmacs-vterm-hook)

```

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -243,8 +243,10 @@ Then triggers a redraw from the module."
 (defun vterm--sentinel (process event)
   "Sentinel of vterm PROCESS."
   (when (not (process-live-p process))
-    (with-current-buffer (process-buffer process)
-      (run-hooks 'vterm-exit-hook))))
+    (let ((buf (process-buffer process)))
+      (when (buffer-live-p buf)
+        (with-current-buffer buf
+          (run-hooks 'vterm-exit-hook))))))
 
 (defun vterm--window-size-change (frame)
   "Callback triggered by a size change of the FRAME.


akermu:
For me this works, can you check again with `emacs -Q`?

jixiuf:
with this 
```
(defun vmacs-vterm-hook()
  (let ((p (get-buffer-process (current-buffer))))
    (when p
      (set-process-query-on-exit-flag p nil))))

(add-hook 'vterm-mode-hook 'vmacs-vterm-hook)
(add-hook 'vterm-exit-hook 'kill-buffer) ;; <------------ with this line  
```

akermu:
I am sorry, but I still cannot reproduce your issue. Could you please give me step for step instructions to produce the error with `emacs -Q`?

jixiuf:
put this in ~/.emacs
 ```
(add-to-list 'load-path "~/repos/emacs-libvterm")
(require 'vterm)

(defun vmacs-vterm-hook()
  (let ((p (get-buffer-process (current-buffer))))
    (when p
      (set-process-query-on-exit-flag p nil))))

(add-hook 'vterm-mode-hook 'vmacs-vterm-hook)
(add-hook 'vterm-exit-hook 'kill-buffer) 

```

1. emacs 

2. M-x:vterm 
3. then C-xk kill this vterm buffer 

then will get 
```
error in process sentinel: save-current-buffer: Selecting deleted buffer
error in process sentinel: Selecting deleted buffer
```

**Comment by akermu** [comment]
Thanks, now I can reproduce the issue! Unfortunately, I really don't like the idea to run the hook conditionally, as a user might want to do other stuff there besides killing the buffer. You could check for the liveliness of the buffer in your exit hook:
```elisp
(add-hook 'vterm-exit-hook (lambda ()
			     (when (buffer-live-p (current-buffer)
			       (kill-buffer))))
```
Is this a satisfying for you? Or do you have another idea to solve this issue?

QiangF:
Using the vterm-exit-hook, if the buffer is killed will produce an error

    error in process sentinel: Selecting deleted buffer

jixiuf:
ok

akermu:
@jixiuf How about giving the buffer as an argument (`nil' if the buffer is deleted) to the `vterm-exit-hook' instead of wrapping it in `with-current-buffer'? That should solve the problem with the error message and still will run the hook unconditionally.

jixiuf:
agree .

PR #35, opened 2018-11-19
Fix auto-install

deb0ch:
With the asynchronous compilation there would always be an error thrown on the `(require 'vterm-module)`, as there is no `vterm-module` before the compilation is done. Using `call-process` instead of `make-process` makes the compilation is synchronous, so the `(require 'vterm-module)` will only be called once it is finished.

There is also another fix: 38eaf28 (`Add CI`) broke the library's output directory and made `emacs-libvterm` unusable as Emacs was not able to find the library anymore.

I need this PR for integrating `emacs-libvterm` properly into Spacemacs (https://github.com/syl20bnr/spacemacs/pull/11552).

I also removed the `(pop-to-buffer buffer)` and `(delete-window window)`. The user doesn't need to see the compilation and it is awkward when you have your window layout and installing vterm breaks it. 

Now the compilation is completely transparent, to the point that in Spacemacs, since we use lazy-loading, the compilation happens the first time the user runs `vterm` and all he sees is a 1-2s delay before the terminal appears, and a "compilation success" message in the echo area.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ set_target_properties(vterm-module PROPERTIES
   C_STANDARD 99
   POSITION_INDEPENDENT_CODE ON
   PREFIX ""
+  LIBRARY_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR}
   )
 
 # Link with libvterm
--- a/vterm.el
+++ b/vterm.el
@@ -7,25 +7,26 @@
 
 ;;; Code:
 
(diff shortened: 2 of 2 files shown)


akermu:
Thanks again! I only changed one thing, namely if an error happens during compilation, then we use `pop-to-buffer` to show the error.

deb0ch:
Yay, thanks ! Indeed using `pop-to-buffer` like that looks better 😸

PR #36, opened 2018-11-19
enable yanking (pasting from clipboard / kill-ring)

deb0ch:
Now you can yank from Emacs and paste from your system's clipboard to your terminal 👌:sparkles: 

I agree that in principle, sending the chars one at a time might not be very optimised, but there is no function yet to send strings to vterm from Emacs, I don't know how to implement one, don't know the edge cases that there will be if I try to, don't really have the time for it, and it works quite well like this for now.

Happy yanking 😁

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -45,7 +45,7 @@
   :type 'number
   :group 'vterm)
 
-(defcustom vterm-keymap-exceptions '("C-c" "C-x" "C-u" "C-g" "C-h" "M-x" "M-o" "C-v" "M-v")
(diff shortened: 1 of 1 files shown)


deb0ch:
Hi @QiangF, I don't understand what you are talking about and what is your point about it, can you elaborate ? There is one what that does what, and what about the gpastel package ?

QiangF:
Sorry, it doesn't apply to vterm. I mean using gpastel and exwm, you can share the clipboard even in xterm.

akermu:
Thanks!

deb0ch:
Yay ! 🎉🎊:sparkles:

PR #39, opened 2018-11-23
Improves performance when receiving large bursts of data.

jixiuf:


jixiuf:
update

deb0ch:
Why would anyone opt-in (or not opt-in) to this ?

jixiuf:
big improvement after last force-pushed  
by setting    ` process-adaptive-read-buffering`
```
(setq process-adaptive-read-buffering  nil)
```

akermu:
Thanks again!

PR #41, opened 2018-11-25
run hook vterm-exit-hook with  the term buffer as argument

jixiuf:
this is related with #30

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -54,7 +54,7 @@ to the terminal anymore."
   :type '(repeat string)
   :group 'vterm)
 
-(defcustom vterm-exit-hook nil
+(defcustom vterm-exit-functions nil
   "Shell exit hook.
 
 This hook applies only to new vterms, created after setting this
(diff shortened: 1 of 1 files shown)


TheBB:
The hok should be renamed to `vterm-exit-functions` to comply with Emacs naming conventions. Hooks don't take arguments.

https://www.gnu.org/software/emacs/manual/html_node/emacs/Hooks.html

> Most hooks are normal hooks. This means that when Emacs runs the hook, it calls each hook function in turn, with no arguments. We have made an effort to keep most hooks normal, so that you can use them in a uniform way. Every variable whose name ends in `-hook` is a normal hook.
> 
> A few hooks are abnormal hooks. Their names end in `-functions`, instead of `-hook` (some old code may also use the deprecated suffix `-hooks`). What makes these hooks abnormal is the way its functions are called—perhaps they are given arguments, or perhaps the values they return are used in some way.

PR #51, opened 2018-12-18
support paste and undo

jixiuf:


akermu:
For me undo works correctly if I bind `C-_` to `vterm-self-insert`. And yanking also works.. @jixiuf Could you explain me, what this PR adds on top?

jixiuf:
> For me undo works correctly if I bind `C-_` to `vterm-self-insert`. And yanking also works.. @jixiuf Could you explain me, what this PR adds on top?

it only works for bash ,not for zsh .

deb0ch:
Also, the way yanking is currently implemented is that elisp code sends the characters one by one to the terminal, which is not the most optimised way to go about it

akermu:
@jixiuf But why does `C-_` work` and `C-/` doesn't?

akermu:
I like to support bracket paste. But to keep the code simple, can we just use the old char by char code and add a special key like `"start_paste"` and `"end_paste"`. Then we don't need the import from glib.

jixiuf:
> I like to support bracket paste. But to keep the code simple, can we just use the old char by char code and add a special key like `"start_paste"` and `"end_paste"`. Then we don't need the import from glib.

done

vterm.el
```diff
@@ -137,6 +137,7 @@ for different shell. "
 (define-key vterm-mode-map (kbd "C-c C-c")             #'vterm-send-ctrl-c)
 (define-key vterm-mode-map (kbd "C-_")                 #'vterm--self-insert)
 (define-key vterm-mode-map (kbd "C-SPC")               #'vterm--self-insert)
+(define-key vterm-mode-map (kbd "C-/")   #'vterm-undo)
```
deb0ch: Should keep it aligned with the rest

vterm.el
```diff
@@ -176,16 +177,26 @@ for different shell. "
   (interactive)
   (vterm-send-key "c" nil nil t))
 
+(defun vterm-undo()
+  (interactive)
+  (vterm-send-key "_" nil nil t))
+
+
```
deb0ch: extra blank line does not match coding style of the file

vterm.el
```diff
@@ -176,16 +177,26 @@ for different shell. "
   (interactive)
   (vterm-send-key "c" nil nil t))
 
+(defun vterm-undo()
```
deb0ch: Space between function name and parameter list. I thought it wouldn't even work without a space in Elisp ?

akermu:
> > I like to support bracket paste. But to keep the code simple, can we just use the old char by char code and add a special key like "start_paste" and "end_paste". Then we don't need the import from glib.
> 
> Does it mean that we will never _ever_ have a fast paste ?

Nope, but I want to try the simple approach first and if it's not sufficient, we can optimize it.

PR #52, opened 2018-12-21
Check vterm-keymap-exceptions for function keys

TheBB:
I don't know if this was intentional or not. I use F12 to toggle the terminal so it was useful for me to exclude that key.


akermu:
No that was not intentional, thanks for the contribution!

PR #64, opened 2019-01-31
Add env INSIDE_EMACS to vterm

suonlight:
# Summary

My zsh config is based on env `INSIDE_EMACS` to alias things like: `vi=emacsclient`


akermu:
Thanks!

Issue #40, opened 2018-11-25
enter/return does not always works


jixiuf:
should similar to https://github.com/vim/vim/issues/1998

in redis-cli or fzf enter/return behavior like Ctrl-j in iterm2

QiangF:
I noticed that with the default setting for fzf, enter doesn't do accept like in xterm, so I have this setting:

    export FZF_DEFAULT_OPTS="--extended --cycle --bind=\
    ctrl-j:accept,ctrl-k:kill-line,tab:toggle-up,btab:toggle-down,\
    ctrl-y:execute('echo {} | xclip -selection clipboard')+abort"

QiangF:
We need this to be solved to user marker. https://github.com/pindexis/marker

deb0ch:
@QiangF You just made me discover Marker and it looks great 😺

QiangF:
@deb0ch You may also like ralph, I am trying to make maker and ralph work together.
https://github.com/qw3rtman/ralph

mattiasb:
> I noticed that with the default setting for fzf, enter doesn't do accept like in xterm, so I have this setting:
> 
> ```
> export FZF_DEFAULT_OPTS="--extended --cycle --bind=\
> ctrl-j:accept,ctrl-k:kill-line,tab:toggle-up,btab:toggle-down,\
> ctrl-y:execute('echo {} | xclip -selection clipboard')+abort"
> ```

For anyone else reading this, I *belive* all other settings above except `--bind=ctrl-j:accept` is noise and not needed to work around this specific issue. So add something like this to your environment and you should be good:
```
FZF_DEFAULT_OPTS=--bind=ctrl-j:accept
```

PR #70, opened 2019-03-15
Fix: enter/return does not always works

suonlight:
# Issue

https://github.com/akermu/emacs-libvterm/issues/40

# Summary

It's similar function `term-send-return` on https://www.emacswiki.org/emacs/download/multi-term.el.
It works well for me on FZF, Tmux inside Vterm.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -169,7 +169,7 @@ for different shell"
 (define-key vterm-mode-map [tab]                       #'vterm--self-insert)
 (define-key vterm-mode-map [backspace]                 #'vterm--self-insert)
 (define-key vterm-mode-map [M-backspace]               #'vterm--self-insert)
-(define-key vterm-mode-map [return]                    #'vterm--self-insert)
+(define-key vterm-mode-map [return]                    #'vterm-send-return)
 (define-key vterm-mode-map [left]                      #'vterm--self-insert)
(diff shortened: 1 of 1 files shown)


QiangF:
I got error in console emacs with this setting. So I changed to

      (when (display-graphic-p)
      (define-key visual-line-mode-map "<return>" #'vterm-send-return))

jixiuf:
`term-send-raw-string` is in term.el ,maybe we can  work around this bug like this , but don't know whether it has some side effects
```
(defun vterm--flush-output (output)
  "Sends the virtual terminal's OUTPUT to the shell."
  (if (and (string-equal output "\C-j")
           (equal this-command 'vterm-send-return))
      (process-send-string vterm--process "\C-m")
    (process-send-string vterm--process output)))
```
```
(define-key vterm-mode-map [return]                    #'vterm-send-return)
(defun vterm-send-return ()
  "Sends C-m to the libvterm."
  (interactive)
  (vterm-send-key "<return>" ))
```

suonlight:
@jixiuf thanks for your code.

suonlight:
@akermu I just rebased the code. It's ready for review.

sadhu89:
Worked for me. Thanks for sharing it!

PR #72, opened 2019-03-24
fix window resize on emacs27

jixiuf:
get this error on emacs 27
```
Error during redisplay: (vterm--window-size-change #<window 13 on ll ~>) signaled (error "Window is on a different frame")
```
and I see 
```
window-size-change-functions is a variable defined in ‘C source code’.
Its value is nil

  This variable’s value is permanent if it is given a local binding.
  This variable may be risky if used as a file-local variable.

Documentation:
Functions called during redisplay when window sizes have changed.
The value should be a list of functions that take one argument.

Functions specified buffer-locally are called for each window showing
the corresponding buffer if and only if that window has been added or
changed its buffer or its total or body size since the last redisplay.
In this case the window is passed as argument.

Functions specified by the default value are called for each frame if
at least one window on that frame has been added or changed its buffer
or its total or body size since the last redisplay.  In this case the
frame is passed as argument.
```

### DIFF
--- a/elisp.c
+++ b/elisp.c
@@ -123,3 +123,7 @@ void set_title(emacs_env *env, emacs_value string) {
 void vterm_invalidate(emacs_env *env) {
   env->funcall(env, Fvterm_invalidate, 0, NULL);
 }
+bool eq(emacs_env *env, emacs_value e1, emacs_value e2) {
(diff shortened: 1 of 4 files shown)


jhacksworth:
Thank you! This works great and fixes a popup window size issue for me when the frame is split into multiple windows.

akermu:
Thanks! I fixed this in `6adcedf' with a little bit simpler implementation.

Issue #58, opened 2019-01-13
Colors


Sbozzolo:
The following screenshot is how `ansi-term` (top left) with `eterm-256colors`, `vterm` (top right) and `xterm` look like on my machine.

![2019-01-12 17 59 16](https://user-images.githubusercontent.com/9167485/51080183-4465bd80-1694-11e9-8448-5887b94c0d47.png)

Also comparing other programs, e.g. `htop`, it seems that `vterm` produces output only with (probably) 8 colors. 
However, `tput colors` reports the number 256, and also other scripts to print the complete palette successfully show all the colors. 

(I am using `zsh`, if it's relevant)

Why is that?

akermu:
Might be because we [set `$TERM` to `xterm`](https://github.com/akermu/emacs-libvterm/blob/a0eb7fe735cd2eef2e1db63620771b636b2f3a75/vterm.el#L230] instead of `xterm-256color`. Can you modify vterm.el and try it out.

In general `emacs-libvterm` supports 256 colors.

Sbozzolo:
Nothing changes when I set `xterm-256color` instead of `xterm`. Also, my xterm sets `TERM` to `xterm`.

akermu:
Mhm currently the module should break, if you encounter a color from the 256 palette. The issue above might be related to the ansi-colors, where you can define 8 normal colors and 8 high intensity colors. The module currently don't differentiate between them, so this might lead to the differences in the output above.

Sbozzolo:
I pulled, recompiled, restarted emacs. This is what I see now:


![2019-01-28 16 41 42](https://user-images.githubusercontent.com/9167485/51873793-c18f6480-231b-11e9-878c-431a69502082.png)

Sbozzolo:
No, I wasn't setting those variables before, nor I was expecting that I had to set them. Shouldn't they default to the values the theme sets?

akermu:
Themes have to support each package individually, if they define their own faces (Have a look at [zenburn](https://github.com/bbatsov/zenburn-emacs/blob/master/zenburn-theme.el).

Anyway with 976c3f2, I think I finally nailed color handling. No more faces, and your themes color theme should be picked up automatically :)

Sbozzolo:
This is what I see now:
![2019-01-29 15 44 04](https://user-images.githubusercontent.com/9167485/51946054-b3098180-23dd-11e9-8009-c2796215f37e.png)

It seems much better, but the background is still off. 
[...]

Issue #56, opened 2019-01-04
How to use for comint and compilation modes?


sandinmyjoints:
Is this even possible? It very well may not be but I wanted to ask just in case. Compilation mode gets sluggish for me when I use it to run large suites of unit tests with spec reporter output. I'd love to use the speed of vterm while still enjoying the niceties of compilation mode like hyperlinked errors, if possible.

brotzeit:
I've also been thinking about this since I'm adding libvterm support to remacs. My goal is to eventually replace the emacs terminal emulation, however I'm expecting difficulties as I probably need to port most term related lisp functions. I didn't take a closer look at it, but I expect that it will be a lot of work. However it should be possible for remacs, I guess.

Regarding emacs-libvterm, I think that's not possible but I really don't know.

jixiuf:
a workaround
```
function compile(){
    tmpfile="/tmp/scratch-`/bin/date +%Y-%m-%d_%H-%M-%S`-`uuidgen`"
     `$@ 2>&1  |cat>$tmpfile &`; emacsclient --no-wait --eval "(with-current-buffer (find-file \"$tmpfile\")(setq default-directory \"`pwd`\")(goto-char (point-max))(compilation-mode) (auto-revert-tail-mode))">/dev/null
}

```
```
compile make test
compile go build 
```

akermu:
Currently it's not possible and I also currently don't know how feasible it is. Adding support for the error "links" on top of vterm might be the way to go..

deb0ch:
> Adding support for the error "links" on top of vterm might be the way to go..

Would require parsing all the output of the shell process in elisp, and we all know how performant that would be...

Or maybe, if there was a way that `vterm-module` would do the parsing in C and somehow give Emacs some metadata about what's in the buffer ?

deb0ch:
In any case, having clickable file names would be an awesome feature

jixiuf:
it seems that  just need to enable `compilation-shell-minor-mode` in vterm-mode-hook 
and then you can call `next-error`

jixiuf:
this should be supported after #126 is merged

azzamsa:
what is the status of `comint` support.

It will be great (at least for me) to use `helm-comint-input-ring` as before.

Thank you

PR #82, opened 2019-04-22
Support 256color

jixiuf:
#58

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,7 @@ if(USE_SYSTEM_LIBVTERM)
 else()
   ExternalProject_add(libvterm
     GIT_REPOSITORY https://github.com/neovim/libvterm.git
-    GIT_TAG a6293a0e033e7e86c74889b4527787993656883a
+    GIT_TAG 89675ffdda615ffc3f29d1c47a933f4f44183364
     CONFIGURE_COMMAND ""
     BUILD_COMMAND make "CFLAGS='-fPIC'"
     BUILD_IN_SOURCE ON
--- a/README.md
+++ b/README.md
@@ -77,22 +77,17 @@ bound to `C-c C-c`.
 Set the `:foreground` and `:background` attributes of the following faces to a
 color you like:
 
-- vterm-color-default-fg
-- vterm-color-default-bg
-- vterm-color-black-fg
-- vterm-color-black-bg
-- vterm-color-red-fg
-- vterm-color-green-bg
-- vterm-color-green-fg
-- vterm-color-yellow-bg
-- vterm-color-blue-fg
-- vterm-color-blue-bg
-- vterm-color-magenta-fg
-- vterm-color-magenta-bg
-- vterm-color-cyan-fg
-- vterm-color-cyan-bg
-- vterm-color-white-fg
-- vterm-color-white-bg
+- vterm-color-default
+- vterm-color-black
+- vterm-color-black
+- vterm-color-red
+- vterm-color-green
+- vterm-color-yellow
+- vterm-color-blue
+- vterm-color-magenta
+- vterm-color-cyan
+- vterm-color-white
+
 
 ## Related packages
 
--- a/elisp.c
+++ b/elisp.c
@@ -113,7 +113,6 @@ emacs_value vterm_get_color(emacs_env *env, int index) {
   emacs_value idx = env->make_integer(env, index);
   return env->funcall(env, Fvterm_get_color, 1, (emacs_value[]){idx});
 }
-
 void set_title(emacs_env *env, emacs_value string) {
   env->funcall(env, Fvterm_set_title, 1, (emacs_value[]){string});
 }
--- a/elisp.c
+++ b/elisp.c
@@ -109,14 +109,9 @@ void set_cursor_type(emacs_env *env, emacs_value QCursorType) {
 }
 
 
-emacs_value get_hex_color_fg(emacs_env *env, emacs_value face) {
-  return env->funcall(env, Fvterm_face_color_hex, 2,
-                      (emacs_value[]){face, Qforeground});
-}
-
-emacs_value get_hex_color_bg(emacs_env *env, emacs_value face) {
-  return env->funcall(env, Fvterm_face_color_hex, 2,
-                      (emacs_value[]){face, Qbackground});
+emacs_value vterm_get_color(emacs_env *env, int index) {
+  emacs_value idx = env->make_integer(env, index);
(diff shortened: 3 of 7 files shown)


suonlight:
It works fine for me. Thank you @jixiuf

Sbozzolo:
I tried this patch. I don't see any changes in the visual appearance (which is still broken with solarized, the theme I am using).

suonlight:
@Sbozzolo: What is your term currently? `echo $TERM` to know that.

Sbozzolo:
`xterm-256colors`

suonlight:
My term is `export TERM=xterm-256color`. Can you remove `s` at the end of your term?

Sbozzolo:
Actually, I miscopied, it is without the "s".

suonlight:
You use `curl -s https://gist.githubusercontent.com/HaleTom/89ffe32783f89f403bba96bd7bcd1263/raw/ | bash` to test 256 colors.

Screenshot 2019-04-23 10 59 01

Sbozzolo:
I see 256 colors, but some of them look different if a run the script with `ansi-term`.

suonlight:
Do you set `ansi colors` like `(setq ansi-color-names-vector ["#292b2d" "#ce527a" "#2d9474" "#bfa325" "#4e97d6" "#bb6dc3" "#299ba2" "#e4e4e4"])`

Screenshot 2019-04-23 11 10 45

suonlight:
I have set the same ansi colors for both emacs and iterm

Sbozzolo:
I don't set anything. I run everything inside emacs. I use eterm-256color
for providing ansi-term with the support for 256 colors.

Anyhow, I was hoping that this PR addressed the issue discussed in #56, but
it seems that it is not the case.

Thanks for the support!

jixiuf:
I think the 8 color in ansi-color-names-vector is not enough 
we need  define two face for default background and default foreground 

https://github.com/akermu/emacs-libvterm/blob/master/vterm-module.c#L460
https://github.com/akermu/emacs-libvterm/blob/master/vterm-module.c#L464
is not correct

maybe we need revert https://github.com/akermu/emacs-libvterm/commit/39d4c44e3b0f2ccf788780c1701cc4b2c90a5d82 and  https://github.com/akermu/emacs-libvterm/commit/976c3f27b5908cedbe7d80ac2f63c00a976e82cd

Sbozzolo:
I checked out the latest commit, and I was extremely happy to see that it works!
I didn't add any of @jixiuf customizations. 

Screenshot to compare XTerm, VTerm and term:

Sbozzolo:
Small problem with `htop`:

(Top pane is vterm, some strings are unreadable. Bottom is xterm)

jixiuf:
can not reproduced with theme color-theme-sanityinc-solarized @Sbozzolo

Sbozzolo:
This is my setting for the theme. I have nothing relevant in the `custom.el`.

```
    (use-package color-theme-sanityinc-solarized
      :config
      (load-theme 'sanityinc-solarized-dark t)
      ;; Flat modeline
      (set-face-attribute 'mode-line nil :box nil)
      (set-face-attribute 'mode-line-inactive nil :box nil)
      ;;(load-theme 'sanityinc-solarized-light t)
      )
```

As you see, I don't do anything peculiar.

jixiuf writes:

jixiuf:
can you try to run `make run ` in emacs-libvterm/build directory

and install color-theme-sanityinc-solarized

Sbozzolo:
I performed some tests. If load emacs with the theme and vterm only, then I get the screenshot above. If I load emacs with vterm, and once emacs is running I load the theme, then the colors are slightly different, and the column name in `htop` are readable.

Do you have suggestions for understanding what is going on?

jixiuf:
trying change the order of loading vterm and  your theme, and see what's going on ,
and trying load your theme in after-init-hook and see what's going on 
```
(add-hook 'after-init-hook (lambda () (load-theme 'solarized-light)))
```

Sbozzolo:
Changing order, or using the lambda you posted changes nothing. 
However, I noticed that if I remove my `.Xresources`, both the scenario (loading in the init or after) yield the same scenario with wrong colors.

akermu:
@jixiuf Is this ready to be reviewed or are you still working on this?

jixiuf:
ready to be reviewed

qhuyduong:
@jixiuf I'm facing this issue with Emacs 26.2. I'm using latest code. Any idea?
```
Error running timer ‘vterm--delayed-redraw’: (void-function vterm--get-color) [2 times]
Error during redisplay: (vterm--window-size-change-26 #<frame vterm<2> – Doom Emacs 0x10614dcb8>) signaled (void-function vterm--get-color)
```

jixiuf:
> @jixiuf I'm facing this issue with Emacs 26.2. I'm using latest code. Any idea?
> 
> ```
> Error running timer ‘vterm--delayed-redraw’: (void-function vterm--get-color) [2 times]
> Error during redisplay: (vterm--window-size-change-26 #<frame vterm<2> – Doom Emacs 0x10614dcb8>) signaled (void-function vterm--get-color)
> ```
@qhuyduong 
are you sure you are using the last version vterm.el in this pull request,
vterm--get-color is defined in vterm.el 
make sure you are loading the right vterm.el

qhuyduong:
@jixiuf my bad. I just build and update vterm-module.so without updating vterm.el. Thank you very much :D

Sbozzolo:
I don't know if the following comment is needed/wanted/welcomed, but:
I've been using this patch on a daily basis for several weeks now.
While the color rendering is still not perfect, on my system it is a massive improvement compared to the out-of-the-box experience.

jixiuf:
after last commit ,the bug should be fixed.
because the color are  inherited  from term.el ,so we should  require term.el .
I don't know  is this a good idea.

or maybe we can create PR to different theme packages for this package instead of require term.el

Sbozzolo:
I checked out the latest commit. There are still imperfections in the rendering. 
For instance, see the attached screenshot, there is an autosuggestion that is barely visible.
What should happen is that the suggested continuation should be lighter compared to the command already typed in, with the previous commit it was of the same weight, as you can see in the second attached screenshot. What I would like to more akin the last screenshot (where it is clear that there is a suggestion).

Btw: thanks for all your contributions to this package, they are invaluable!

jixiuf:
I compare vterm  and iterm2 ,and found they both are imperfections in the rendering ,like the first  screenshot you attached. so I thinks it should be the problem of the theme.
Btw: is the third screenshot  using solarized theme, and I found the colors are different with screenshot1 and screenshot2.

jixiuf:
rebase the code.

Issue #98, opened 2019-07-08
Blinking cursor by default


sutram:
Great package. I have been using it regularly.

I think after https://github.com/akermu/emacs-libvterm/pull/88 was merged, I noticed that the cursor in my emacs session has started blinking.

Is it possible to change that behavior back to not blink by default?

Right now, as a workaround, I have to add (blink-cursor-mode 0) explicitly to the end of my init file.

Thanks.

jixiuf:
because emacs enable `blink-cursor-mode` by default .

sutram:
@jixiuf Thanks for responding. I am using emacs 26.2 and before the merge, the cursor would not blink. I haven't upgraded emacs before or after the merge.

Thanks.

akermu:
@sutram Maybe something changed in your configuration? You can disable cursor blinking by putting `(blink-cursor-mode -1)` in your file?

Yevgnen:
Even having `(blink-cursor-mode -1)` in `init.el`, starting a `vterm` invalid the settings.

Since `blink-cursor-mode` is a global minor mode, I personally think packages or libraries should not change user settings.

jubnzv:
I personally used the following solution for a quick fix:

```emacs-lisp
(defun jbz:vterm-restore-blink (buffer)
  "Disable blink-cursor-mode. Somehow vterm enables it on start."
  (blink-cursor-mode 0))

(add-to-list 'vterm-exit-functions #'jbz:vterm-restore-blink)
```

sutram:
@jubnzv Thank you! I'll try your workaround. The blinking cursor is driving me crazy. Even adding (blink-cursor-mode 0) in the init file was not working after emacs session restore so I had to also add  blink-cursor-blinks 1 to the init file.

jixiuf:
```

(define-minor-mode blink-cursor-mode
  "Toggle cursor blinking (Blink Cursor mode).

If the value of `blink-cursor-blinks' is positive (10 by default),
the cursor stops blinking after that number of blinks, if Emacs
gets no input during that time.

See also `blink-cursor-interval' and `blink-cursor-delay'.

This command is effective only on graphical frames.  On text-only
terminals, cursor blinking is controlled by the terminal."
  :init-value (not (or noninteractive
		       no-blinking-cursor
		       (eq system-type 'ms-dos)
		       (not (display-blink-cursor-p))))
  :initialize 'custom-initialize-delay
  :group 'cursor
  :global t
[...]

PR #104, opened 2019-07-26
Revert "trying support blink-cursor-mode"

jixiuf:
blink-cursor-mode is a global minor-mode ,and there is no way to   enable or disable blink-cursor-mode only for vterm buffer .  #98


deb0ch:
It might be a little late (catching up on months of discussions), but I see you write 

> blink-cursor-mode is a global minor-mode ,and there is no way to enable or disable blink-cursor-mode only for vterm buffer

and actually, **there _is_ a way of inhibiting a global minor mode in specific (major mode) buffers**, [see how I did it for centered-cursor-mode and vterm](https://github.com/syl20bnr/spacemacs/pull/11552/commits/0b3e96c011c8268e925e871bd008132bcf807ef8#diff-5faa6bac0457b3d8431901cff894666bR212).

It may seem a bit hackish, but the Stackoverflow answer linked in the comments contains detailed explanation of why this way works and others won't (I tried several).

Maybe this will help for future support ?

Issue #55, opened 2019-01-02
Update default-directory


deadtrickster:
Is it possible to update `default-directory` after cd so I can easily find files? At least title updated somehow...

deadtrickster:
This is how I hacked it for now: 

```elisp
(defun vterm--rename-buffer-as-title (title)
  (let ((dir (concat (nth 1 (split-string title ":")) "/")))
    (cd dir))
  (rename-buffer (format "term %s" title)))
(add-hook 'vterm-set-title-functions 'vterm--rename-buffer-as-title)
```

title is <user>@<host>:path

deb0ch:
I saw a PR and a few commits about that recently, they called it "directory tracking". I think you can find it in the recently closed PRs. If I recall correctly you have to setup some stuff in your shell's config files as well for it to work.

QiangF:
The PR is https://github.com/akermu/emacs-libvterm/pull/29

jhacksworth:
@deadtrickster Thanks for this! Now `C-x C-f` works the way it should. 😃 

For me, removing the buffer renaming and just updating the directory allows for multiple vterm buffers to have the same current directory.

```elisp
(defun vterm--cd-to-directory-in-title (title)
  "Change into directory extracted from path info in TITLE.

The format of TITLE is expected to be '<user>@<host>:path'

TITLE is based on the xterm style window title, set via the shell. See sections
3 and 4 on this page for how to set the TITLE via the shell:
http://www.faqs.org/docs/Linux-mini/Xterm-Title.html"
  (let ((dir (concat (nth 1 (split-string title ":")) "/")))
    (cd dir)))
(add-hook 'vterm-set-title-functions #'vterm--cd-to-directory-in-title)
```

QiangF:
I think leaving cwd in the buffer name helps a lot in finding the right vterm.

jixiuf:
> For me, removing the buffer renaming and just updating the directory allows for multiple vterm buffers to have the same current directory.
```

(rename-buffer NEWNAME &optional UNIQUE)

:around advice: ‘uniquify--rename-buffer-advice’

Change current buffer’s name to NEWNAME (a string).
If second arg UNIQUE is nil or omitted, it is an error if a
buffer named NEWNAME already exists.
If UNIQUE is non-nil, come up with a new name using
‘generate-new-buffer-name’.
Interactively, you can set UNIQUE with a prefix argument.
We return the name we actually gave the buffer.
[...]

PR #112, opened 2019-08-03
Add remote directory tracking

akermu:
#55

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,16 @@ cd() {
 }
 ```
 
+## Remote directory tracking
+
+Put this in your *remote* .zshrc:
+
+```zsh
+function chpwd() {
+    print -Pn "\e]51;$(whoami)@$(hostname):$(pwd)\e\\"
+}
+```
+
 ## Related packages
 
 - [vterm-toggle](https://github.com/jixiuf/vterm-toggle): Toggles between a vterm and the current buffer
--- a/vterm.el
+++ b/vterm.el
@@ -531,8 +531,19 @@ Feeds the size change to the virtual terminal."
 
 (defun vterm--set-directory (path)
   "Set `default-directory' to PATH."
-  (when (file-directory-p path)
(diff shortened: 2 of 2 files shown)


akermu:
Another solution would be to guess the user and domain from the terminal via a regexp. Then the user wouldn't have to modify their remote `.zshrc` or `.bashrc`.

jixiuf:
> Another solution would be to guess the user and domain from the terminal via a regexp (probably via `tramp-shell-prompt-pattern`). Then the user wouldn't have to modify their remote `.zshrc` or `.bashrc`.

how ? I don't think we can  always guess the user,host and directory   ,different people may set different prompt .


and how about define some local variable like 
```
(defvar-local vterm-user "")
(defvar-local vterm-host "")
(defvar-local vterm-dir "")
```
and run a hook when directory changed.

jixiuf:
Change my mind ,I approved it  ,let's keep it simple .

jixiuf:
how about we change 
```
    print -Pn "\e]51;$(pwd)\e\\";
```
to 
```
    print -Pn "\e]51;A$(pwd)\e\\";
```
the difference is there is a `A` after `51;`
in case we add other osc_callback later
https://www.iterm2.com/documentation-escape-codes.html
like iterm use `^[]1337;` as the prefix of  non-standard escape codes

akermu:
> how about we change
> 
> ```
>     print -Pn "\e]51;$(pwd)\e\\";
> ```
> 
> to
> 
> ```
>     print -Pn "\e]51;A$(pwd)\e\\";
> ```
> 
> the difference is there is a `A` after `51;`
> in case we add other osc_callback later
> https://www.iterm2.com/documentation-escape-codes.html
> like iterm use `^[]1337;` as the prefix of non-standard escape codes

This is a good idea, but `51` would be our `1337` or do I miss something?

jixiuf:
yes ,51 would be our 1337

PR #119, opened 2019-08-07
Save and restore point in copy mode.

mbrumlow:
This change will save the point when entering copy mode and restore it back once
leaving. This feels much nicer than having the point get out of sync with where
the terminal actually thinks it is.

If the creators of vterm know of a better way I would be glad to figure out how
to implement it, but for now this seems to work fine.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -298,7 +298,10 @@ If nil, never delay")
   (if vterm-copy-mode
       (progn                            ;enable vterm-copy-mode
         (use-local-map nil)
-        (vterm-send-stop))
+        (vterm-send-stop)
+        (set (make-local-variable 'vterm-copy-saved-point) (point)))
+    (if vterm-copy-saved-point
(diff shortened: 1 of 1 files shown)


vterm.el
```diff
@@ -298,7 +298,10 @@ If nil, never delay")
   (if vterm-copy-mode
       (progn                            ;enable vterm-copy-mode
         (use-local-map nil)
-        (vterm-send-stop))
+        (vterm-send-stop)
+        (set (make-local-variable 'vterm-copy-saved-point) (point)))
```
jixiuf: I think you should define a local variable  like `vterm--process` 
and I suggest you rename  `vterm-copy-saved-point` to `vterm--copy-saved-point`

mbrumlow: Thank, I will do that shortly.

mbrumlow: All updated with your suggestions, thanks!

PR #122, opened 2019-08-08
change directory tracking prefix form 51; to 51;A

jixiuf:
https://github.com/akermu/emacs-libvterm/pull/112#issuecomment-519561616

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ For `zsh` put this in your `.zshrc`:
 
 ```zsh
 function chpwd() {
-    print -Pn "\e]51;$(pwd)\e\\";
+    print -Pn "\e]51;A$(pwd)\e\\";
 }
 ```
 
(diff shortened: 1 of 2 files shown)


akermu:
So just for my understanding, what is the additional A for? Why is 51; not enough?

Or does 51 denote the "application ID" and "A" the command ID?

jixiuf:
> Or does 51 denote the "application ID" and "A" the command ID?
yes
if we want add other  non-standard escape codes
then we can easily add another `51;B` or `51;C` 

https://www.iterm2.com/documentation-escape-codes.html
see the `Shell Integration/FinalTerm` section 
```
Escape Sequences
FinalTerm originally defined various escape sequences in its original spec that are not supported by iTerm2 and are not described in this document. The best remaining references to these codes are in iTerm2's source code.

FTCS_PROMPT

^[]133;A^G

Sent just before start of shell prompt.

FTCS_COMMAND_START

^[]133;B^G

Sent just after end of shell prompt, before the user-entered command.

FTCS_COMMAND_EXECUTED

^[]133;C^G
```
I think our `51;` equals to there `133;` here

PR #126, opened 2019-08-09
Support compilation shell minor mode

jixiuf:
#56 
vterm enable `compilation-shell-minor-mode` by default.
even you change your working directory, `next-error` `previous-error` still
can find out the corresponding source code.

For `zsh` put this in your `.zshrc`:

```zsh
vterm_prompt() {
    print -Pn "\e]51;A$(pwd)\e\\";
}
PROMPT=$PROMPT'%{$(vterm_prompt)%}'

```
please use `PROMPT`, and do NOT use `chpwd` or `preexec`,
unless you don't want to using `compilation-shell-minor-mode`.
The working directory of lines after a new prompt  need be updated .

For bash 

```bash
vterm_prompt_end(){
  printf "\e]51;A$(pwd)\e\\"
}
PS1=$PS1'$(vterm_prompt_end)'
```

this is ready to be reviewed ,but need more testing before merge.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -76,6 +76,7 @@
 (require 'subr-x)
 (require 'cl-lib)
 (require 'color)
+(require 'compile)
 
 (defcustom vterm-shell shell-file-name
   "The shell that gets run in the vterm."
@@ -261,7 +262,35 @@ If nil, never delay")
            :sentinel (when vterm-exit-functions #'vterm--sentinel))))
   (vterm--set-pty-name vterm--term (process-tty-name vterm--process))
   (process-put vterm--process 'adjust-window-size-function
-               #'vterm--window-adjust-process-window-size))
+               #'vterm--window-adjust-process-window-size)
+  (setq next-error-function 'vterm-next-error-function))
+
+(defun vterm--compilation-setup ()
+  "function to setup `compilation-shell-minor-mode' for vterm.
+`'compilation-shell-minor-mode' would change the value of
+local variable `next-error-function',so we should call this function in
+`compilation-shell-minor-mode-hook'."
+  (when (eq major-mode 'vterm-mode)
+    (setq next-error-function 'vterm-next-error-function)))
+
+(add-hook 'compilation-shell-minor-mode-hook #'vterm--compilation-setup)
+
+;;;###autoload
+(defun vterm-next-error-function (n &optional reset)
+  "Advance to the next error message and visit the file where the error was.
+This is the value of `next-error-function' in Compilation buffers."
+  (interactive "p")
+  (let* ((pt (point))
+         (msg (compilation-next-error (or n 1) nil
+				                      (or compilation-current-error
+					                      compilation-messages-start
+					                      (point-min))))
+         (default-directory default-directory)
+         (pwd (vterm--get-pwd)))
+    (when pwd
+      (setq default-directory pwd))
+    (goto-char pt)
+    (compilation-next-error-function n reset)))
 
 ;; Function keys and most of C- and M- bindings
 (defun vterm--exclude-keys (exceptions)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -99,32 +98,50 @@ color you like:
...
+can find out the corresponding source code.
+
+
 For `zsh` put this in your `.zshrc`:
 
 ```zsh
-function chpwd() {
-    print -Pn "\e]51;A$(pwd)\e\\";
+vterm_prompt() {
+      print -Pn "\e]51;A$(pwd)\007"
```
akermu: Why did you change this to `\007`?

README.md
```diff
@@ -99,32 +98,50 @@ color you like:
...
 ## Directory tracking
 
+The `default-directory` would change when you change your working directory.
+and text in the vterm buffer has a text property `working-directory`.
+
+
+
+vterm enable `compilation-shell-minor-mode` by default.
+even you change your working directory, `next-error` `previous-error` still
+can find out the corresponding source code.
```
akermu: This has to be reformulated to something like

```With `compilation-shell-minor-mode` enabled, you can jump to file references via `next-error` and `previous-error` like in a `compilation-mode` buffer.```
The technical details don't belong into the README.

File: README.md [hunk 3]
```diff
@@ -99,32 +98,50 @@ color you like:
...
-    print -Pn "\e]51;A$(pwd)\e\\";
+vterm_prompt() {
+      print -Pn "\e]51;A$(pwd)\007"
 }
+PROMPT=$PROMPT'%{$(vterm_prompt)%}'
+
 ```
+please use `PROMPT`, and do NOT use `chpwd` or `preexec`,
+unless you don't want to using `compilation-shell-minor-mode`.
+The `working-directory` text property need be updated when a new shell prompt come out.
```
akermu: Pleas use proper casing, e.g. `Please`, make the NOT bold instead of Uppercase. Maybe formulate like this:
```If you want to use `compilation-shell-minor-mode`, please use `PROMPT` and **not** `chpwd` or `preexec`.

File: README.md [hunk 4]
```diff
@@ -99,32 +98,50 @@ color you like:
...
+```
 
 ```zsh
-function chpwd() {
-    print -Pn "\e]51A;$(whoami)@$(hostname):$(pwd)\e\\"
+vterm_prompt() {
+    print -Pn "\e]51;A$(whoami)@$(hostname):$(pwd)\e\\"
 }
+PROMPT=$PROMPT'%{$(vterm_prompt)%}'
+# I need get the `pwd` just when I got a prompt ,so don't use chpwd() or preexec()
```
akermu: Please remove this line.

vterm.el
```diff
@@ -529,29 +549,33 @@ Feeds the size change to the virtual terminal."
...
-                  (setq default-directory dir)))
-            (setq default-directory (concat "/-:" path)))))
-    (when (file-directory-p path)
-      (setq default-directory path))))
+  (let ((dir (vterm--get-directory path)))
+    (when dir (setq default-directory dir))))
+
+(defun vterm--get-directory (path)
+  "Get  normalized directory to PATH."
+  (when path
```
akermu: This should not be needed, does it?

README.md
jixiuf: just copy and modified from my original zshrc config, I will change it to `\e\\`

vterm.el
jixiuf: this is used in `vterm-next-error-function`

jixiuf:
> I hadn't had the time to review the C code proper, but will do it in the next days.
> 
> I have a general question: Could the directory tracking not only be implemented as a C function which returns the directory for a given line, instead of a text-property?
> 
`vterm--redraw` is called by a timer ,I worried it would have some problem.
I will try to do that if it has no problem.
> Also a general remark: I really like that your doing some refactoring work along the way 👍 But it makes reviewing a little hard, because I have to figure out what belongs to the feature and what belongs to the refactoring. So if you don't mind, split the PR up next time, thanks!
I removed  some unrelated code

akermu:
There was also commit in there, which removed the `count_lines` elisp call, could you make a seperate PR for that?

jixiuf:
> There was also commit in there, which removed the `count_lines` elisp call, could you make a seperate PR for that?

ok

jixiuf:
> I hadn't had the time to review the C code proper, but will do it in the next days.
> 
> I have a general question: Could the directory tracking not only be implemented as a C function which returns the directory for a given line, instead of a text-property?

done.

and should  ready to be reviewed

Sbozzolo:
`zsh` offers the function `precmd`, which is executed before each prompt. I think that using this function is more elegant than changing directly the prompt. My suggestion, following #137 (which, by the way, probably comes from a stackoverflow answer), is
``` sh
HOSTNAME="$(hostname)" # or whatever is needed to connect to the machine
USER="$(whoami)"
sync_dir() {
    print -Pn "\e]51;A${USER}@${HOSTNAME}:$(pwd)\e\\"
}
precmd_functions=($precmd_functions sync_dir)
```
for remote shells, or
``` sh
sync_dir() {
    print -Pn "\e]51;A$(pwd)\e\\"
}
precmd_functions=($precmd_functions sync_dir)
```
for local ones.

jixiuf:
#146 is related .
in #146 I want to get the position of the end of  each  prompt, so I use the `PROMPT` 
I want  the callback can be  executed **after** each prompt instead of **before** .

README.md
```diff
@@ -102,32 +101,45 @@ is ansi color 8-15.
...
+If you want to use `compilation-shell-minor-mode`, please use `PROMPT` and not `chpwd` or `preexec`.
+
+For bash 
 
 ```bash
-cd() {
-  builtin cd "$@" || return
-  [ "$OLDPWD" = "$PWD" ] || echo -e "\e]51;A$(pwd)\e\\"
-}
+PS1='\$ \[\e]51;A$(pwd)\e\\\]'
```
benley: Since PS1 always has _some_ value set, you can suggest appending to it instead of replacing the existing value:

```suggestion
PS1+='\[\e]51;A$(pwd)\e\\\]'
```

I was about to also suggest using `\w` instead of `$(pwd)` since it's going in PS1, but it turns out that can break if a user sets `PROMPT_DIRTRIM`, so I supose `$(pwd)` is better.

PR #129, opened 2019-08-11
support display-line-numbers-mode.

jixiuf:
width should = window-width - line-num-width

~~don't known why `echo $COLUMNS` still  equals to `window-body-width`~~
ready to be reviewed

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -494,8 +494,8 @@ Argument BUFFER the terminal buffer."
             (inhibit-read-only t))
         (when vterm--term
           (when (and (require 'display-line-numbers nil 'noerror)
-                     (get-buffer-window)
-                     (display-line-numbers-update-width))
+                     (get-buffer-window buffer t)
+                     (ignore-errors (display-line-numbers-update-width)))
             (window--adjust-process-windows))
           (vterm--redraw vterm--term)))
       (setq vterm--redraw-timer nil))))
--- a/vterm.el
+++ b/vterm.el
@@ -224,18 +224,14 @@ If nil, never delay")
   "Major mode for vterm buffer."
   (buffer-disable-undo)
   (setq vterm--term (vterm--new (window-body-height)
-                                (window-body-width)
+                                (- (window-body-width) (vterm--get-margin-width))
                                 vterm-max-scrollback))
 
   (setq buffer-read-only t)
   (setq-local scroll-conservatively 101)
   (setq-local scroll-margin 0)
-
-  (if (version< emacs-version "27")
-      (add-hook 'window-size-change-functions #'vterm--window-size-change-26 t t)
-    (add-hook 'window-size-change-functions #'vterm--window-size-change t t))
   (let ((process-environment (append `(,(concat "TERM="
-						vterm-term-environment-variable)
+						                        vterm-term-environment-variable)
                                        "INSIDE_EMACS=vterm"
                                        "LINES"
                                        "COLUMNS")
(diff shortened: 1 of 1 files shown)


akermu:
This is a emacs bug since the description of `window-body-width` says it returns the width of the text area. Do you agree?

jixiuf:
yeah ,agree. but don't known when would they fix this? and would they consider it as a bug?

akermu:
Apparently is is not considered as abug: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=37020

jixiuf:
the width of  line number would change and we can't hook it when it change

akermu:
We could use the `adjust-window-size-function` property of our process, which is called [here](https://github.com/emacs-mirror/emacs/blob/405f851f4bf64e2290e841b65ffabf37c61187f4/lisp/window.el#L9829).

jixiuf:
yeah ,but I think this function is only called  when the window size change.
It won't be called when the width of line number changed .

and in this pull request,I use `window-adjust-process-window-size-function`,
I think it has no difference with `adjust-window-size-function` property of our process

jixiuf:
maybe add an advice for `window-adjust-process-window-size-function` is not a good idea,
I can change it to use `adjust-window-size-function` .,

vterm.el
```diff
@@ -490,26 +488,31 @@ Argument EVENT process event."
...
+  (let* ((size (funcall window-adjust-process-window-size-function
+                        process windows))
+         (width (car size))
+         (height (cdr size))
+         (inhibit-read-only t))
+    (setq width (- width (vterm--get-margin-width)))
+    (when (and (processp process)
+               (process-live-p process))
+      (vterm--set-size vterm--term height width))
+    (cons width  height)))
```
akermu: Please remove the double space.

vterm.el
```diff
@@ -490,26 +488,31 @@ Argument EVENT process event."
...
-Feeds the size change to the virtual terminal."
-  (with-current-buffer (window-buffer window)
-    (when (and (processp vterm--process)
-               (process-live-p vterm--process))
-      (let ((height (window-body-height window))
-            (width (window-body-width window))
-            (inhibit-read-only t))
-        (set-process-window-size vterm--process height width)
-        (vterm--set-size vterm--term height width)))))
+(defun vterm--window-adjust-process-window-size(process windows)
```
akermu: Why not use `display-line-numbers-update-width`. The cursor should be anyway at the end of the buffer. We should just think about, how it interacts with `vterm-copy-mode`.

vterm.el
```diff
@@ -490,26 +492,27 @@ Argument EVENT process event."
...
+               (process-live-p process))
+      (vterm--set-size vterm--term height width))
+    (cons width height)))
+
+(defun vterm--get-margin-width ()
+  "Get margin width of vterm buffer when `display-line-numbers-mode' is enabled."
+  (let ((width 0))
+    (when (and (boundp 'display-line-numbers)
+               display-line-numbers)
+      (require 'display-line-numbers)
```
akermu: This should not be needed, since `display-line-numbers` is bound and set here.

jixiuf: I think this is needed.

 `display-line-numbers` is defined in C from emacs26.1.

1.  I think we should still support emacs <26.1.
2. One can just enable display line numbers by `(setq display-line-numbers t)` in a mode hook without require display-line-numbers.el

jixiuf:
I have test this for about month, so I thinks it is stable.

PR #146, opened 2019-08-18
[WIP] Shell Integration  (mark up a shell's output with semantic information )

jixiuf:
not finished yet.
and   should be merge after #126 .
there are some refactoring in commit `refactoring  term->dirs and add LineInfo.`  related to #126 
EDIT:
 Want to mark up a shell's output with semantic information about where the prompt ends, where the user-entered command ends.

with `vterm-next-prompt`  and `vterm-previous-prompt`  you can jump to the next/previous prompt position .

And I also want to save the killed line to kill ring when I press `C-k' in the vterm buffer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,12 @@ is ansi color 8-15.
 - vterm-color-cyan
 - vterm-color-white
 
-## Directory tracking
+## Shell Integration & Directory tracking
+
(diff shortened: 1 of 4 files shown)


akermu:
Could you give an short explanation on what this does?

PR #150, opened 2019-08-22
Updated directory tracking snippet for zsh

muffinmad:
Check if `INSIDE_EMACS` is `vterm` because this code make sense only in vterm.
Also added trailing slash.


jixiuf:
`INSIDE_EMACS` does not work on remote server .

Issue #137, opened 2019-08-14
Remote directory tracking without `cd`


Sbozzolo:
At the moment, at least for `zsh`, the suggested way to let `vterm` know which directory we are visiting is with the `chpwd` hook. As a result, when logging in into a remote machine, the directory-tracking doesn't work immediately, but it starts working only after the first `cd`. Clearly, this is a minor problem, but it is still an annoyance. 

Is there any reason why the current solution is preferable to the following, in which the problem doesn't appear?
``` sh
HOSTNAME="$(hostname)" # or whatever is needed to connect to the machine
USER="$(whoami)"
sync_dir() {
    print -Pn "\e]51;A${USER}@${HOSTNAME}:$(pwd)\e\\"
}
precmd_functions=($precmd_functions sync_dir)
```

As a side remark, the current suggestion assumes that hostname provides the correct way to connect to the machine. However, this is not necessarily the case. The lines above, while more verbose, are clearer in this respect.

akermu:
Hi I'm happy to accept an PR for this. The other solution is to put it in the prompt as suggested here: https://github.com/akermu/emacs-libvterm/pull/126

jixiuf:
this should be fixed

PR #173, opened 2019-10-22
update readme about directory tracking.

jixiuf:
#137
 and  supporting directory tracking for fish


Sbozzolo:
What is the reason for this change? For example, this
```zsh
 print -Pn "\e]51;A$(whoami)@$(hostname):$(pwd)\e\\";
```
works also on localhost, so most likely the if statement is not needed.

PR #178, opened 2019-10-29
Fix shift-space behavior

r-zip:
This PR improves handling of shift-space. I find that I often hold down shift after typing special characters, like `!`. Using `emacs-libvterm`, <kbd>S-SPC</kbd> currently sends an escape sequence that ends up being displayed on the line.

With the changes in this PR, holding shift will send a space normally, like in other terminal programs. macOS's Terminal.app and iTerm2 both do this, so it seems like standard behavior.


Sbozzolo:
Thanks. This is the same behavior as in the built-in `term` or `shell`.

Sbozzolo:
I've used this patch for 17 days without any problem. Can it be merged, or is there a better way to address the problem?

PR #189, opened 2019-11-30
Some convenience commands

casouri:
Here are some convenience functions that I personally use. Maybe they can be helpful.

Oops, I just saw there is a vterm-toggle, please ignore the second commit then.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -722,5 +722,75 @@ the called functions."
         (apply (cadr f) args)
       (message "Failed to find command: %s" command))))
 
+;;; Convenience
+
+(defun vterm--rename-buffer-as-title (title)
+  "Rename buffer name as the xterm TITLE."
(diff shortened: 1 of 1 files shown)


akirak:
Thank you for this PR. Some of the features look useful, so I'm looking forward to receive them soon.

There is a package named `vterm-toggle` on MELPA, which provides a command named `vterm-toggle`. I don't think you have to remove your command with the same name, but it would be better renaming it to avoid the conflict.

casouri:
I wrote my toggle function before knowing the existence of that package. From a glance, it is much better than mine, so I don't think the second commit, which contains the code about my toggle function) is needed anymore. It can be either cherry-picked or I can remove that commit and do a force push.

jixiuf:
I cherry-picked your merge request.  @casouri

PR #191, opened 2019-12-01
Separate elisp code that builds external dependency and the rest of elisp code

akater:
Currently, on each update vterm recompiles the module disregarding the general system settings and also fetches libvterm. This is very annoyng and requires integration of vterm into a system package manager.

When installing from an external package manager, it makes no sense to even have `vterm-module-compile` available unless it is kept aligned with system-wide settings. Keeping it aligned, however, essentially requires patching the package twice: patching command string in elisp, and introducing the same options on a more general level, which is a very unwelcome redundancy.

It's best if elisp build infrastructure can be disregarded completely; for that, `vterm.el` has to be split and all build-related stuff should go to its own elisp module. It is also beneficial due to the fact that external package manager may use emacs to build `vterm-module` without loading any other elisp files.

I isolated the make function, modified its parameters string so that configuring patches, if any, look better (an option per line), replaced `(eq 0)` with `zerop`, and eliminated a variable that does not bring anything to the table. I also explicitly mentioned the `files` dependency because I usually do it.

`vterm-module-make` might nevertheless still need some commented preamble; I'm not aware of what's considered good style here.

### DIFF
--- a/vterm-module-make.el
+++ b/vterm-module-make.el
@@ -1,7 +1,6 @@
 ;;; vterm-module-make.el --- make vterm-module in elisp  -*- lexical-binding: t; -*-
 
 (require 'files)
-(require 'term)
 
 (defvar vterm-install-buffer-name " *Install vterm"
   "Name of the buffer used for compiling vterm-module.")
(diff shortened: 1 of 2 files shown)


akirak:
> vterm-module-make might nevertheless still need some commented preamble; I'm not aware of what's considered good style here.

As described in [this document](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org), you should run `checkdoc` and `package-lint` to check the elisp source file.

Issue #179, opened 2019-11-08
re-wrap line when window resized


kenkangxgwe:
Currently, when you change the windows width, 
- __enlarge width__: the line wrapped at the original width.
    _actual_:
    ```
    -------------------------      --------------------------------------
    | $This is a long sente |      | $This is a long sente              |
    | nce that got wrapped  |  ->  | nce that got wrapped               |
    | before resize         |      | before resize                      | 
    -------------------------      --------------------------------------
    ```
   __expected_:
    ```
    -------------------------      --------------------------------------
    | $This is a long sente |      | $This is a long sentence that got  |
    | nce that got wrapped  |  ->  | wrapped before resize              |
    | before resize         |      |                                    | 
    -------------------------      --------------------------------------
    ```
- __shrink width__: the line got truncated at new width.
    _actual_:
    ```
    -------------------------      --------------------
    | $This is a long sente |      | $This is a long s|
    | nce that got wrapped  |  ->  | nce that got wrap|
    | before resize         |      | before resize    | 
    -------------------------      --------------------
    ```
    _expected_:
    ```
    -------------------------      --------------------
    | $This is a long sente |      | $This is a long  |
    | nce that got wrapped  |  ->  | sentence that go |
    | before resize         |      | t wrapped before |
    -------------------------      | resize           |
                                   --------------------
    ```
I wonder if it is easy to do the re-wrapping of the text?

freesteph:
I've ran into the same issue but would be perfectly happy with a no-wrapping, no-truncating version. i.e:

```
+--------------------------------------------------+
| $ some program                                   |
| > a very long output that should not wrap to avoi|d the resize issues
+--------------------------------------------------+
```

Which means I can peep into that buffer full-width/height if I need to grasp a potentially verbose stacktrace and reversely, if I don't, it can use the unbound hidden width to spread, rather than polluting a narrow vterm buffer. 

I've tried Emacs's `toggle-truncate-lines` and `toggle-word-wrap` and no combination achieves it.

plattfot:
I also ran into this issue when testing out `vterm`. `vterm-min-window-width` kind of works around the issue if I set to a large enough number and turn on `toggle-truncate-lines`. But that messes up the output for commands like `ls` and the right sided prompt in `zsh`.

[...]

PR #210, opened 2020-01-11
add defcustom vterm-min-window-width.

jixiuf:
#179


johnmwu:
This isn't observed when initially opening a vterm. When I `M-x vterm` and run `$ echo $COLUMNS`, I get 77, when the minimum is set to 80.

jixiuf:
@johnmwu  should fixed.

PR #240, opened 2020-02-15
[WIP] Ask which libvterm to use upon compilation

Sbozzolo:
`vterm-module-make`  ignores the (possible) existence of a system `libvterm`. Some users may prefer to use the `libvterm` provided by their distribution. With this PR, users will be asked which `libvterm` they want to use. 

This PR requires more tests before being merged.

### DIFF
--- a/vterm-module-make.el
+++ b/vterm-module-make.el
@@ -2,7 +2,7 @@
 
 (require 'files)
 
-(defvar vterm-install-buffer-name " *Install vterm"
+(defvar vterm-install-buffer-name "*Install vterm*"
   "Name of the buffer used for compiling vterm-module.")
 
 (defcustom vterm-compile-with-system-libvterm nil
--- a/vterm-module-make.el
+++ b/vterm-module-make.el
@@ -27,7 +27,10 @@ the system libvterm is used."
                vterm-compile-with-system-libvterm
                (y-or-n-p
                 "Do you want to use the system libvterm? (It has to be installed separately)"))
-              "yes"
+              (progn
+                (setq vterm-compile-with-system-libvterm t) ; If the user says "yes",
+                                                            ; we save their preference
+               "yes")
               "no"))
          (make-commands
           (concat
--- a/vterm-module-make.el
+++ b/vterm-module-make.el
@@ -18,7 +18,7 @@ and a local copy from GitHub."
 If the variable vterm-compile-with-system-libvterm is set to t,
 the system libvterm is used."
   (interactive)
-  (let* ((default-directory
+  (let* ((vterm-directory
            (file-name-directory (file-truename (locate-library "vterm"))))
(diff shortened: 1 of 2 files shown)


vterm-module-make.el
```diff
@@ -2,22 +2,44 @@
...
+               (y-or-n-p
+                "Do you want to use the system libvterm? (It has to be installed separately)"))
+              (progn
+                (setq vterm-compile-with-system-libvterm t) ; If the user says "yes",
+                                                            ; we save their preference
+               "yes")
+              "no"))
+         (make-commands
+          (concat
+           "cd " vterm-directory ";"
```
purcell: Here you definitely need `shell-quote-argument`, in case Emacs is installed into a directory whose name contains special shell chars like spaces or semicolons.

vterm-module-make.el
```diff
@@ -2,22 +2,44 @@
...
-            -DCMAKE_BUILD_TYPE=RelWithDebInfo \
-            ..; \
-          make"))
+  (let* ((vterm-directory
+           (file-name-directory (file-truename (locate-library "vterm"))))
+         ;; Ask the user if they want to compile with the version of libvterm
+         ;; on the system. It has to be available! (We don't perform checks)
+         (compile-with-system-libvterm
+          (if (or
+               vterm-compile-with-system-libvterm
```
purcell: I feel like some people will inevitably get prompted every time they install/update the package if we use a binary option like this, and that would quickly get annoying. I'd suggest that the var should be ternary, with an `'ask` option which triggers prompting. The default would either be `ask`, or an option that works in most cases (as the default currently does nicely for me).

vterm-module-make.el
```diff
@@ -2,22 +2,44 @@
...
+                (setq vterm-compile-with-system-libvterm t) ; If the user says "yes",
+                                                            ; we save their preference
+               "yes")
+              "no"))
+         (make-commands
+          (concat
+           "cd " vterm-directory ";"
+           "mkdir -p build;"
+           "cd build;"
+           "cmake" " -DUSE_SYSTEM_LIBVTERM=" compile-with-system-libvterm
```
purcell: Also good practice to use `shell-quote-argument` here, plus the combination of spaces inside quoted strings and separate strings is a bit weird. ie. why not `"cmake -DUSE_SYSTEM_LIBVTERM="` here?

fbergroth:
How about making the args to cmake a defcustom, with some common options? This gives a bit more flexibility to the user as well.

Sbozzolo:
> How about making the args to cmake a defcustom, with some common options? This gives a bit more flexibility to the user as well.

Yes, I agree. In the next revision of this PR, I will implement this and @purcell's suggestions.

(Thanks @purcell for your precious comments!)

PR #251, opened 2020-02-21
Fix some packaging issues

fbergroth:
Hi, I'm using this package as a submodule (using borg package manager rather than installing via package.el) and found some issues. See each commit.

- Do not compile vterm when loading vterm-module-make
I don't think compilation should run when just loading vterm-module-make. Explicitly calling compile from vterm.el makes more sense.

- Always compile when calling vterm-module-compile
If you have an outdated vterm-module.so, you'll want to be able to recompile it. `(require 'vterm-module nil t)` is already essentially the same check. I want to call vterm-module-compile whenever I update this submodule, but currently it does nothing if vterm-module.so exists.

Then I went loose and fixed byte compilation:
- Add declarations to quiet byte compiler
- Remove unused code
- Inline vterm-bind-key (not needed, but this was an odd macro - am I missing something?)
- Evaluate parts during compilation

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -46,10 +46,11 @@
   (error "VTerm needs module support. Please compile your Emacs
   with the --with-modules option!"))
 
-(unless (require 'vterm-module nil t)
-  (require 'vterm-module-make)
-  (vterm-module-compile)
-  (require 'vterm-module))
+(eval-and-compile
+  (unless (require 'vterm-module nil t)
+    (require 'vterm-module-make)
+    (vterm-module-compile)
+    (require 'vterm-module)))
 
 (declare-function display-line-numbers-update-width "display-line-numbers")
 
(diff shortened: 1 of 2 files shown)


Sbozzolo:
Thanks, @fbergroth, for your contribution. I am re-working a little bit `vterm-module-make` to be more robust, so most likely there will be soon something that will have to be adjusted in your PR.

Sbozzolo:
My plan is to merge #253 (that checks if `cmake` and `libtool` are available), and #240 (when it is finished), before looking at this PR.

vterm.el
```diff
@@ -301,40 +305,32 @@ variable `next-error-function', so we should call this function in
...
-(mapc (lambda (key)
-        (eval `(vterm-define-key ,key)))
-      (cl-loop for prefix in '("C-" "M-")
-               append (cl-loop for char from ?a to ?z
-                               for key = (format "%s%c" prefix char)
-                               collect key)))
+
+(eval-and-compile
+  (defmacro vterm-define-key (key)
+    "Define an command sending M-[a-z] or C-[a-z] to vterm."
```
Sbozzolo: Can you please fix the spelling mistake in this docstring?
"an command" -> "a command"

Feel free to expand upon/rewrite the docstring.

vterm.el
```diff
@@ -301,40 +305,32 @@ variable `next-error-function', so we should call this function in
...
+    (declare (indent defun)
+             (doc-string 3))
+    `(defun ,(intern(format "vterm-send-%s" key))()
```
Sbozzolo: Here, can you insert a blank space after `intern`?

vterm.el
```diff
@@ -688,7 +685,7 @@ Argument EVENT process event."
         (kill-buffer buf))
     ))
 
-(defun vterm--text-scale-mode (&optional argv)
+(defun vterm--text-scale-mode (&optional _argv)
   "fix `line-number' height for scaled text"
```
Sbozzolo: I know there's a lot of work to be done on docstrings (in general in the package)... Can you capitalize the "fix" word?

Sbozzolo:
Thank you, @fbergroth. Sorry for keeping this PR sitting here. All the comments are very minor, once they are addressed, we can merge it.

fbergroth:
@Sbozzolo I'm not touching the lines you requested to be changed, other than indentation. Click the gear icon and "Hide whitespace changes" on the diff page, and it'll be more clear.

The main thing I'm unsure about here is building the module on byte-compilation. The alternative is to forward-declare all variables from the module, which is more cumbersome. I was planning to add makem.sh for linting as well, but never got around to it.

Issue #162, opened 2019-09-12
Font resizing not rendered properly


jhacksworth:
The vterm buffer does not reflect changes in font size properly.

For example, starting with the following screenshot--default font size, running `ls` in a directory. No problems, so far so good:

<img width="853" alt="original-size" src="https://user-images.githubusercontent.com/286057/64825472-6305b680-d58b-11e9-81b1-0fb2a0fcd957.png">

Then, I decrease the font size using the Emacs [command](https://www.gnu.org/software/emacs/manual/html_node/emacs/Text-Scale.html) `text-scale-decrease` and type `ls` again. The output does not reflect the change in font size. There should be more columns in the output of `ls` and the output should continue lower in the window.

<img width="853" alt="resize-font" src="https://user-images.githubusercontent.com/286057/64825503-82044880-d58b-11e9-8ee0-83249df210f6.png">

If I further resize the window and run `ls`, there is more confusion:

<img width="880" alt="resize-window" src="https://user-images.githubusercontent.com/286057/64825599-d90a1d80-d58b-11e9-9e33-11fa5b28023c.png">

This is on MacOS Mojave and Emacs 27.

It would be great to be able to have the vterm font size smaller than the default font sometimes, but even so, vterm is still awesome. Thanks to everyone who has contributed to this package. :)

Sbozzolo:
#230 should (partially) fix this. If you follow the steps you described, you should see that the number of columns of `ls` changes (for new instances of `ls`). The number of columns of past `ls` instances does not change, but the size of the font is properly increased/decreased. Can you check that you find the same behavior? Does this close the issue?

hexmode:
My default font is a proportional font, but I change it for certain modes by using the following code:
``` emacs
(defun mah/set-current-buffer-fixed ()
  "Set the current buffer to use fixed pitch."
  (interactive)
  (set (make-local-variable 'buffer-face-mode-face) 'fixed-pitch)
  (buffer-face-mode t))

;; Fix up these modes for some of the global settings I have
(defvar mah/fixed-width-modes
  '("prog" "org-agenda" "shell" "term" "diff" "dired" "calendar"
    "ssh-tunnels" "magit-popup" "help" "magit-log" "apache" "sql"
    "picture" "ibuffer" "calc-keypad" "proced" "yaml" "makefile"
    "makefile-gmake" " *transient*" "conf" "vterm"))
(dolist (mode mah/fixed-width-modes)
[...]

PR #252, opened 2020-02-22
Remove special default font and just use the default

hexmode:
See #162 for rationale.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,20 @@ or remove it from `vterm-mode-map`. By default, `vterm.el` binds most of the
 like `<backspace>` and `<return>`. Sending a keyboard interrupt is bound to `C-c
 C-c`.
 
+## Fonts
+
+If you would like to change the font or face used in a vterm, use the following code:
+
+``` emacs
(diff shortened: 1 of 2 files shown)


hexmode:
Oh, and, if it matters, I have a copyright assignment on file with the FSF.

Sbozzolo:
Dear @hexmode, thanks for your contribution.

Your commit essentially reverts back 38bbd46745917f5d9e4062b94b729d92382de9d7, which granted users the possibility of choosing a different font for vterm. I don't fully grasp what is the problem with that, so I'll let @jixiuf (author of 38bbd46745917f5d9e4062b94b729d92382de9d7) take care of this PR.

hexmode:
@jixiuf, @Sbozzolo:  I would suggest that the method I've given in #162 is the "right" way to set a default font for a mode.  At least, it seems to work for most other things.

jixiuf:
can you mention that in readme about set a different font in vterm

hexmode:
@jixiuf does that look good?

Sbozzolo:
Thank you, @hexmode.

PR #253, opened 2020-02-22
Thorw errors if required dependencies are not available

Sbozzolo:


fbergroth:
Some of these errors should be reported during compilation, is it really necessary to reimplement the checks in elisp? With this you have the minimum cmake version declared in two places.

I think it's better to just provide a link to a FAQ in the readme when the compilation fails.

Sbozzolo:
Hi @fbergroth, thanks for your comment. 

I agree that errors should pop up during compilation, but the main reason for this change is to provide a very user-friendly way to report errors. Instead of having a random error appearing during the compilation stage, we clearly inform the user that some required packages are not installed. No need to understand what is going on with the compilation.

Also, I don't see where else we check for the version of cmake. I suspect that if one tries to compile with cmake < 3.11.0, there would be some error from cmake itself.

> I think it's better to just provide a link to a FAQ in the readme when the compilation fails.

Ideally, I would like this to be somewhere in the docstrings/commentary of the package too.

Please, continue sharing your opinions, they are valuable ways to improve the package.

fbergroth:
> I agree that errors should pop up during compilation, but the main reason for this change is to provide a very user-friendly way to report errors. Instead of having a random error appearing during the compilation stage, we clearly inform the user that some required packages are not installed. No need to understand what is going on with the compilation.

I agree, but I think the error should come from the build tool and not from elisp wrapping the build tool. You can check stuff and generate errors from cmake.

> Also, I don't see where else we check for the version of cmake. I suspect that if one tries to compile with cmake < 3.11.0, there would be some error from cmake itself.

It's here: https://github.com/akermu/emacs-libvterm/blob/61a5be82c9ebd3a153c4801e992d6287a0fa2e87/CMakeLists.txt#L1
A message like
```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.11 or higher is required.  You are running version X
```
will be displayed if the user has a lower version.

Sbozzolo:
> I agree, but I think the error should come from the build tool and not from elisp wrapping the build tool.
Why is that? 

We know very clearly which executables `vterm` requires, and we know that it doesn't make sense to try compiling if those are not available. My main goal with this change is to make the error less intimidating. Users that will find these errors will know that they have to install something. Users that found the old error may be clueless (if they are less experienced). I don't see what is the drawback of this approach.

> It's here: [...]

You are right. I forgot about that. I still stand by my opinion for the reasons described above.

fbergroth:
My point is that cmake is the build tool, and is the canonical place to declare build dependencies. Otherwise the build tool will become a cmake+elisp hybrid, which is unnecessary.

I agree that checking if cmake is available is a good idea, but checking its version already works, and does not need to be duplicated.

Regarding libtool, I think this is the proper way to do it in cmake (I'm no cmake expert):
```
find_program(LIBTOOL NAMES libtool glibtool)
if(NOT LIBTOOL)
  message(FATAL_ERROR "libtool must be installed")
endif()
```
I don't really see how errors are intimidating if done like this.

Also, I think that maybe libtool check is only necessary if `USE_SYSTEM_LIBVTERM=false`, i.e. in the branch with ExternalProject_add.

vterm-module-make.el
```diff
@@ -5,11 +5,54 @@
...
+  ;;
+  ;; So, we extract the version from the first line.
+  (let* ((cmake-min-required-version "3.11.0")
+        (cmake-version-output (shell-command-to-string "cmake --version"))
+        (cmake-version
+         (progn
+           (string-match "^cmake version \\([0-9.]+\\)" cmake-version-output)
+           (match-string 1 cmake-version-output))))
+    (when
+        (string< cmake-version cmake-min-required-version)
```
fbergroth: Just a note, use `version<` when comparing versions. `(string< "3.0.0" "10.0.0")` is false.

Sbozzolo: > My point is that cmake is the build tool, and is the canonical place to declare build dependencies. Otherwise the build tool will become a cmake+elisp hybrid, which is unnecessary.
> 
> I agree that checking if cmake is available is a good idea, but checking its version already works, and does not need to be duplicated.
> 
> Regarding libtool, I think this is the proper way to do it in cmake (I'm no cmake expert):
> 
> ```
> find_program(LIBTOOL NAMES libtool glibtool)
> if(NOT LIBTOOL)
>   message(FATAL_ERROR "libtool must be installed")
> endif()
> ```
> 
> I don't really see how errors are intimidating if done like this.
> 
> Also, I think that maybe libtool check is only necessary if `USE_SYSTEM_LIBVTERM=false`, i.e. in the branch with ExternalProject_add.

Thanks, I am happy with your suggestions. I think this is a good way to handle this.

PR #257, opened 2020-02-27
vterm-copy-mode-map bind vterm-bol.

jixiuf:


Sbozzolo:
I pulled and quickly tried this, but it doesn't work on my system. I think vterm cannot identify properly my prompt. I'll debug this in the weekend

vterm.el
```diff
@@ -848,6 +849,19 @@ the called functions."
          (or (= pt prompt-pt)
              (string-blank-p (buffer-substring-no-properties pt prompt-pt))))))
 
+(defun vterm-bol ()
```
Sbozzolo: I think it would be better to call this function `vterm-beginning-of-line`, following Emacs' own `beginning-of-line`.

Then, I'd modify the docstring with
```emacs-lisp
"Move point to the beginning of the line.

Move the point to the first character after the shell prompt on this line.
If the point is already there, move to the beginning of the line.
Effectively toggle between the two positions.
```

Then, I still don't understand exactly how the prompt is detected... I think it would be useful to explain this in the README and in some docstrings (if it requires setup from the shell side).

Sbozzolo:
I checked again this, and if I found that it works if I put all the relevant configurations in my `.zshrc`. It would be great if we could have more documentation on the detection of the prompt.

vterm.el
jixiuf: I renamed `vterm-bol` to `vterm-beginning-of-line`, And docstring is updated .

jixiuf:
> I checked again this, and if I found that it works if I put all the relevant configurations in my `.zshrc`. It would be great if we could have more documentation on the detection of the prompt.

```
vterm-prompt-end() { vterm-printf "51;A$(whoami)@$(hostname):$(pwd)"; }
PROMPT=$PROMPT"%{$(vterm-prompt-end)%}"
```
https://github.com/akermu/emacs-libvterm/blob/master/vterm-module.c#L870

Issue #243, opened 2020-02-16
Auto-compilation of module changes working dir of vterm


purcell:
Loving `vterm`, thanks! I noticed a quirk which affects me each time the version is bumped:

### Steps to reproduce:

* Install a new version of `vterm`
* Run `M-x vterm`

### Expected result

* Module is compiled
* `vterm` session is started with the default directory of the buffer from which it launched

### Actual result

* Module is compiled
* `vterm` session starts with the working directory set to the installation dir of the new package, e.g. `~.emacs.d/elpa/vterm-20200210.704`

purcell:
I should note that I looked at the code and it wasn't immediately obvious why this should happen.

Sbozzolo:
In `vterm-module-make`, we set `default-directory` to the folder where `vterm.el` was downloaded. I don't fully understand what is going with the binding of this variable, but it seems that this is why you see the reported behavior. 

I am working on an improved version of `vterm-module-make` that allows the user to choose which libvterm to use (the system one, or the one downloaded from git). This is in PR #240. In the latest commit, I removed all the references to `default-directory`, and added a `cd` to the compilation command. As a result, my tests indicate that `vterm` now behaves as expected.

The commit history in #240 is messy (most commits are buggy due to a missing blank space - I'll clean it up before I merge it), but the last commit should be fine regardless the version of libvterm you want to use.

purcell:
> I don't fully understand what is going with the binding of this variable, but it seems that this is why you see the reported behavior.

Thanks, I left a few comments on #240.

Sbozzolo:
PR #261 implements a workaround that, according to my tests, fixes this problem.

PR #261, opened 2020-03-02
Update build system

Sbozzolo:
This PR implements the improvements discussed in #253 and #240. Namely:
1. By default, vterm uses libvterm pulled from GitHub, now updated to 0.1.3
2. Users can change this behavior by customizing the variable `vterm-module-compilation-flags`
3. We check that `cmake` is available before compiling. If it is not available, we throw an error
4. We check that libtool is available (during compile-time), if not, we inform the user
5. We implement a workaround to avoid changing the `default-directory`, "fixing" #243 

If this PR is merged, #240 and #253 will be closed (without being merged).

@fbergroth, are you interested in reviewing this PR?

### DIFF
--- a/vterm-module-make.el
+++ b/vterm-module-make.el
@@ -2,7 +2,7 @@
 
 (require 'files)
 
-(defcustom vterm-module-custom-compilation-flags ""
+(defcustom vterm-module-compilation-flags ""
   "Flags given to CMake to compile vterm-module.
 
 Currently, the flags available are:
@@ -43,7 +43,7 @@ the executable."
              cd build; \
              cmake \
              -DCMAKE_BUILD_TYPE=RelWithDebInfo "
-            vterm-module-custom-compilation-flags
+            vterm-module-compilation-flags
             " ..; \
              make; \
              cd -"
--- a/vterm-module-make.el
+++ b/vterm-module-make.el
@@ -2,7 +2,7 @@
 
 (require 'files)
 
-(defvar vterm-install-buffer-name " *Install vterm"
+(defvar vterm-install-buffer-name " *Install vterm* "
   "Name of the buffer used for compiling vterm-module.")
 
 (defun vterm-module--cmake-is-available ()
--- a/vterm-module-make.el
+++ b/vterm-module-make.el
@@ -2,10 +2,11 @@
 
 (require 'files)
 
-(defcustom vterm-module-compilation-flags ""
-  "Flags given to CMake to compile vterm-module.
+(defcustom vterm-module-cmake-args ""
+  "Arguments given to CMake to compile vterm-module.
 
-Currently, the flags available are:
+Currently, vterm defines the following flags (in addition to the
+ones already available in CMake):
 
 `USE_SYSTEM_LIBVTERM'. Set it to `yes' to use the version of
 libvterm installed on your system.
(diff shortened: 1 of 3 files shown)


fbergroth:
> @fbergroth, are you interested in reviewing this PR?

Sure!

- Perhaps make the libtool check just before ExternalProject_add?
- It would be nice to get rid of the hardcoded `-DCMAKE_BUILD_TYPE=RelWithDebInfo` (so users have the option to set another build type with `vterm-module-compilation-flags`). You can do something like this to set the default in cmake (I haven't tested this):
  ```
  if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
    message(STATUS "No build type selected, defaulting to RelWithDebInfo")
    set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Build type (default RelWithDebInfo)" FORCE)
  endif()
  ```
  Another similar recipe [here](https://cliutils.gitlab.io/modern-cmake/chapters/features.html). I can also add that in #251 if you wish.

Rest is looking good :+1:

vterm-module-make.el
```diff
@@ -2,28 +2,58 @@
...
-      (let* ((buffer (get-buffer-create vterm-install-buffer-name)))
-        (pop-to-buffer vterm-install-buffer-name)
-        (if (zerop (call-process "sh" nil buffer t "-c" make-commands))
-            (message "Compilation of emacs-libvterm module succeeded")
-          (error "Compilation of emacs-libvterm module failed!"))))))
+  (when (vterm-module--cmake-is-available)
+    (let* ((vterm-directory
+            (shell-quote-argument
+             (file-name-directory (file-truename (locate-library "vterm")))))
+          (make-commands
```
fbergroth: This is not properly indented

vterm-module-make.el
```diff
@@ -2,28 +2,58 @@
...
+            "cd " vterm-directory "; \
+             mkdir -p build; \
+             cd build; \
+             cmake \
+             -DCMAKE_BUILD_TYPE=RelWithDebInfo "
+            vterm-module-compilation-flags
+            " ..; \
+             make; \
+             cd -"
+            )))
```
fbergroth: Join the trailing parens to the previous line

vterm-module-make.el
Sbozzolo: This is how `emacs-lisp-mode` indents this block. How would you indent it?

vterm-module-make.el
```diff
@@ -2,28 +2,56 @@
...
+          (make-commands
+           (concat
+            "cd " vterm-directory "; \
+             mkdir -p build; \
+             cd build; \
+             cmake "
+             vterm-module-compilation-flags
+            " ..; \
+             make; \
+             cd -")))
```
fbergroth:
```suggestion
           (make-commands
            (concat
             "cd " vterm-directory "; \
             mkdir -p build; \
             cd build; \
             cmake "
             vterm-module-compilation-flags
             " ..; \
             make; \
             cd -")))
```

vterm-module-make.el
fbergroth: See https://github.com/akermu/emacs-libvterm/pull/261#discussion_r386833022. `make-commands` is not on the same column as `vterm-directory` in the let binding.

Sbozzolo: > make-commands is not on the same column as vterm-directory in the let binding. 

Yes, I see it now.

Sbozzolo:
Thanks, @fbergroth. I addressed your comments.

fbergroth:
Regarding the docstring of `vterm-module-compilation-flags`, there are a lot more options. You can get all available options with `cmake .. -LA` (from the build directory). Otherwise LGTM.

PR #274, opened 2020-03-20
This commit fixes the newlines when copied and pasted.

mbrumlow:
This commit fixes up the newlines that were inserted for wrapping using the same
approach as term.el does. It does so by adding a property to the '\n' chars that
were inserted as a result of wrapping the terminal. When killing a region text
will be passed through the filter-buffer-substring-function
vterm--remove-fake-newlines removing the wrapped new lines by looking for the
'vterm-line-wrap property and removing the char that corresponds.

If this change set is not up to project standards please work with me to correct
any issues you find.

### DIFF
--- a/vterm-module.c
+++ b/vterm-module.c
@@ -258,7 +258,7 @@ static void refresh_lines(Term *term, emacs_env *env, int start_row,
   int offset = 0;
   for (i = start_row; i < end_row; i++) {
 
-    int newline = 0; 
+    int newline = 0;
     for (j = 0; j < end_col; j++) {
       fetch_cell(term, i, j, &cell);
 
@@ -295,15 +295,13 @@ static void refresh_lines(Term *term, emacs_env *env, int start_row,
       }
     }
 
-
-    if(!newline) { 
+    if (!newline) {
       emacs_value text = render_text(env, term, buffer, length, &lastCell);
       insert(env, text);
       length = 0;
       text = render_fake_newline(env, term, buffer, length, &lastCell);
       insert(env, text);
     }
-    
   }
   emacs_value text = render_text(env, term, buffer, length, &lastCell);
   insert(env, text);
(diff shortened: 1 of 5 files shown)


jixiuf:
Can you make it work with `vterm-copy-mode`.
When `vterm-copy-mode` is enabled, the faked newline is removed.
And when `vterm-copy-mode` is disabled, the faked newline is recovered.

also can you run 
```
clang-format -i *.h
clang-format -i *.c 
```

mbrumlow:
Actually removing the newlines in the render is a lot more involved.  With the way this module has been built I suspect it would require re-rendering the entire buffer when entering copy mode -- this could cause some performance problems.  I also have not yet wrapped my head around cursor position account. I am currently working on efforts for re-flowing the text on resize. Once that happens we may be in a better position to dynamically change the newlines when entering copy mode.  Prior to this I was able to let emacs do the wrapping but cursor position calculations did not work right :( 

This change really only focuses on them being removed from anything  killed/copied.

mbrumlow:
```
clang-format -i *.h
clang-format -i *.c 
```
Has been ran.

vterm-module.c
```diff
@@ -657,6 +669,25 @@ static emacs_value render_text(emacs_env *env, Term *term, char *buffer,
   return text;
 }
 
+static emacs_value render_fake_newline(emacs_env *env, Term *term, char *buffer,
+                                       int len, VTermScreenCell *cell) {
```
jixiuf: Can you remove unused arguments here.

jixiuf:
Sorry, my last `force-pushed`  is pushed by accident. just ignore it .

vterm-module.c
mbrumlow: Done.

Sbozzolo:
Can you please add some documentation/comments in the code too?

PR #275, opened 2020-03-20
Adding documentation.

mbrumlow:
The other PR was merged just after the request for some documentation. This is me adding that documentation.

### DIFF
--- a/vterm-module.c
+++ b/vterm-module.c
@@ -667,6 +667,10 @@ static emacs_value render_text(emacs_env *env, Term *term, char *buffer,
   return text;
 }
 
+// render_fake_newline attaches a text property to the newline being injected by
+// the terminal emulator when it reaches edge of the terminal. This text
+// property is later used to identify such insertions and remove them when the
+// text is being killed/copied, thus preserving the output's natural line breaks.
 static emacs_value render_fake_newline(emacs_env *env, Term *term) {
 
   emacs_value text;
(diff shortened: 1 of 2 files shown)


Sbozzolo:
Thanks, @mbrumlow. This is fantastic.

Can I ask you to expand a little bit on the docstring in the elisp function?
Maybe you can explain what the fake newlines are (ie, why they are there. You can maybe copy from the comments in the C code).

This is to make sure that users can find more documentation from within Emacs.

Sbozzolo:
Thanks, @mbrumlow.

A few comments. There are some guidelines for writing docstrings: https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html

In this case, the first line should be shorter, descriptive and should be imperative. Arguments should appear in the docstring and should be capitalized. 

It is also helpful to crosslink different dosctrings. For instance, the docstring for `vterm--filter-buffer-substring` is not very informative. Linking it to `vterm--remove-fake-newlines` would give the reader a pointer.

Also, neither of the docstrings really explain what are these fake new lines. Where is that they are tagged? Why do we need them?

Sbozzolo:
@mbrumlow, it would be great if you implemented what discussed in my previous message. 
Are you still interested in doing that?

Thanks

mbrumlow:
Yes, I will be working on this -- work has been busy, once I get some free time back I will be sure to address this issue.

PR #276, opened 2020-03-21
reimplements vterm--get-prompt-point with text property and bind vterm-next/previous-prompt

jixiuf:
just like `term-next-prompt` and `term-previous-prompt` in `term.el`
```
(define-key vterm-copy-mode-map (kbd "C-c C-n")        #'vterm-next-prompt)
(define-key vterm-copy-mode-map (kbd "C-c C-p")        #'vterm-previous-prompt)

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,8 +128,8 @@ The package can be installed with `guix package -i emacs-vterm`.
 ## Shell-side configuration
 
 Some of the most useful features in `vterm` (e.g.,
-[directory-tracking](#directory-tracking) or [message
-passing](#message-passing)) require shell-side configurations. The main goal of
+[directory-tracking and prompt-tracking](#directory-tracking-and-prompt-tracking) or
+[message passing](#message-passing)) require shell-side configurations. The main goal of
 these additional functions is to enable the shell to send information to `vterm`
 via properly escaped sequences. A function that helps in this task,
 `vterm_printf`, is defined below. This function is widely used throughout this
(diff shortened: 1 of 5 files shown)


Sbozzolo:
Thanks. I tested and it works very well. Some quick comments:
1. Can you add some documentation? For example, it would be great if you could explain in the docstring of `vterm--get-prompt-point` how the detection of the prompt works.
2. What is `vterm--prompt-enabled-p`?

mbrumlow:
Am I right that libvterm is actually tracking the prompt already? If so this is huge. No more regex madness!  I think we will be able to do a lot of neat things with setting properties on text. 

(I checked this code out and it seemed to work for me)

jixiuf:
> Am I right that libvterm is actually tracking the prompt already? If so this is huge. No more regex madness! I think we will be able to do a lot of neat things with setting properties on text.
> 
> (I checked this code out and it seemed to work for me)

yes.

jixiuf:
> Thanks. I tested and it works very well. Some quick comments:
> 
> 1. Can you add some documentation? For example, it would be great if you could explain in the docstring of `vterm--get-prompt-point` how the detection of the prompt works.

done. forgive my pool English , correct me if it is not right.

> 2. What is `vterm--prompt-enabled-p`?

documentation about vterm--prompt-enabled-p is added.

> 3. Why next/previous-prompt are only in VTermCopy mode?

Ok, I bind them in vterm-mode-map

README.md
```diff
@@ -285,14 +285,26 @@ is ansi color 8-15.
...
-## Directory tracking
+## Directory tracking and Prompt tracking
 
 `vterm` supports _directory tracking_. If this feature is enabled, the default
 directory in Emacs and the current working directory in `vterm` are synced. As a
 result, interactive functions that ask for a path or a file (e.g., `dired` or
 `find-file`) will do so starting from the current location.
 
-Directory tracking requires some configuration, as the shell has to be
+And `vterm` supports _prompt tracking_. If this feature is enabled, Emacs known
```
mbrumlow: known -> knows

jixiuf: fixed

mbrumlow:
*NOTE* On further inspection this is not working. 

vterm--get-prompt-point seems to pass all the way through to term-skip-prompt.

vterm.el
```diff
@@ -831,29 +836,67 @@ the called functions."
...
+More information see `vterm--prompt-tracking-enabled-p' and
+`Directory tracking and Prompt tracking'in README. "
   (let (pt)
     (save-excursion
-      (setq pt (vterm--get-prompt-point-internal
-                vterm--term (line-number-at-pos))))
+      (end-of-line)
+      (if (get-text-property (point) 'vterm-prompt)
+          (setq pt (point))
+        (setq pt (previous-single-property-change (point) 'vterm-prompt))
```
mbrumlow: This `previous-single-property-change` return nil every time :( 

Either vterm-prompt is not getting set in the c code, or libvterm is not actually tracking.

vterm-module.c
```diff
@@ -216,7 +216,22 @@ static bool is_eol(Term *term, int end_col, int row, int col) {
   }
   return 1;
 }
-
+static int is_end_of_prompt(Term *term, int end_col, int row, int col) {
+  LineInfo *info = get_lineinfo(term, row);
+  if (info == NULL) {
+    return 0;
+  }
+  if (info->prompt_col < 0) {
```
mbrumlow: It seems this LineInfo is defined in vterm-module.h and never actually set, or used (until now). So I think the supporting code for tracking the prompt is missing,

mbrumlow: Oh, I am wrong! It is set, but it is expecting OCS callback to help it!

jixiuf: see https://github.com/akermu/emacs-libvterm#directory-tracking

Directory tracking and Prompt tracking requires some configuration, as the shell has to be
instructed to share the relevant information with Emacs.  
For `zsh`, put this at the end of your `.zshrc`:

```zsh
vterm_prompt_end() {
    vterm_printf "51;A$(whoami)@$(hostname):$(pwd)";
}
setopt PROMPT_SUBST
PROMPT=$PROMPT'%{$(vterm_prompt_end)%}'
```

mbrumlow:
Ignore my comments, I missed the step of setting my bash prompt up properly, this is working as expected now that I have corrected my error.

PR #278, opened 2020-04-01
Don't resize while in copy mode.

mbrumlow:
- This change prevents the region and point location from being lost on resize,
and a whole slew of other weird behaviors.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -995,6 +995,11 @@ Argument EVENT process event."
   "Adjust width of window WINDOWS associated to process PROCESS.
 
 `vterm-min-window-width' determines the minimum width allowed."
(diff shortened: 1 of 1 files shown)


Sbozzolo:
What happens if you resize in VTermCopy mode and then go back to 'normal' mode?

mbrumlow:
I use EXWM, so I guess I would not see how this would work. 

I will see about saving that a resize happened in copy mode, and when we exit copy mode handle the resize.

Sbozzolo:
Sorry, @mbrumlow, I didn't understand what your PR was about. I rebased against master, added a comment, and I will merge it.

Fixes: #413

PR #279, opened 2020-04-02
vterm-copy-mode-done copies the line if there is no region

zallison:
Improvements:
- Easier to copy a full line
- Consistent way of exiting vterm-copy-mode
- Still works the same on regions

How the code works:
- If the region is not set then make the current line the region and proceeds from there.
- save-excursion puts the cursor back where it was after setting the region

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -433,13 +433,16 @@ This is the value of `next-error-function' in Compilation buffers."
     (vterm-send-start)))
 
 (defun vterm-copy-mode-done ()
-  "Save the active region to the kill ring and exit `vterm-copy-mode'."
+  "Save the active region or line to the kill ring and exit `vterm-copy-mode'."
   (interactive)
   (unless vterm-copy-mode
     (user-error "This command is effective only in vterm-copy-mode"))
-  (unless (region-active-p)
-    (user-error "No region is active"))
-  (kill-ring-save (region-beginning) (region-end))
+  (save-excursion
+    (unless (region-active-p)
+      (beginning-of-line)
+      (set-mark (point))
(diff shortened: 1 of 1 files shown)


Sbozzolo:
Thanks for your contribution and the clear PR message.

It would be nice to give users the choice of what to do in lines with the prompt. I want more often to copy a command I executed, as opposed to the whole line (including the prompt). 

We could add an option `vterm-copy-include-prompt` that controls the default behavior. 

Also, could you expand the docstring to describe the behavior of the function, and update the README?

zallison:
Given how many custom prompts are out there I'm a bit worried in how that would work.  I suppose we could have an option for `vterm-end-of-prompt` set to `$ ` (or something) by default?  That sounds reasonable.

The docstring in this PR added "or line":
  "Save the active region or line to the kill ring and exit `vterm-copy-mode'."

Does it need more than that?

I'll update the README and see what I can do about prompt options.

zallison:
didn't mean to close it!

Sbozzolo:
Sorry, I should have given you more prompts (pun not intended).

With the most recent commits, we are able to identify users' prompts. A good place where to look is `vterm-beginning-of-line`. In fact, you should be able to easily base your function on that.

> The docstring in this PR added "or line":
> "Save the active region or line to the kill ring and exit `vterm-copy-mode'."

Yes, the docstring doesn't fully describe what the function does. Mainly, it would be great to describe when the function kills the line, and when it kills the region. A user should be able to discover features just by reading the docstring. Same with the README.

zallison:
Thanks for the pointers.  I'll see what I can do.

PR #281, opened 2020-04-03
vterm-copy-mode-done copies the line if there is no region

zallison:
Improvements:

- Consistent way of exiting vterm-copy-mode
- Still works the same on regions
- Easier to copy a full line
- Excluding the prompt from the line copy is possible with or without vterm prompt tracking working.
- With the prefix command invert the excluding of the prompt.

How the code works:

If the region is set it copies that region into the `kill-ring`.

Otherwise it copies from the start of the line or the end of the prompt, as determined by the options `vterm-copy-exclude-prompt`, `vterm-copy-use-vterm-prompt`, and if the latter is nil `vterm-copy-prompt-regexp`.

Cons:
Adds three custom variables and increases the code complexity slightly.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -430,6 +430,7 @@ This is the value of `next-error-function' in Compilation buffers."
 (define-key vterm-copy-mode-map (kbd "RET")            #'vterm-copy-mode-done)
 (define-key vterm-copy-mode-map (kbd "C-c C-r")        #'vterm-reset-cursor-point)
 (define-key vterm-copy-mode-map (kbd "C-a")            #'vterm-beginning-of-line)
+(define-key vterm-copy-mode-map (kbd "C-e")            #'vterm-end-of-line)
 (define-key vterm-copy-mode-map (kbd "C-c C-n")        #'vterm-next-prompt)
 (define-key vterm-copy-mode-map (kbd "C-c C-p")        #'vterm-previous-prompt)
 
@@ -911,6 +912,15 @@ in README."
                  (when pt (goto-char (1- pt)))))
     (term-previous-prompt n)))
 
+(defun vterm--get-end-of-line ()
+  "Find the start of the line, bypassing line wraps."
+  (save-excursion
+    (end-of-line)
+    (while (get-text-property (point) 'vterm-line-wrap)
+      (forward-char)
+      (end-of-line))
+    (point)))
+
 (defun vterm--get-prompt-point ()
   "Get the position of the end of current prompt.
 More information see `vterm--prompt-tracking-enabled-p' and
@@ -956,6 +966,11 @@ Effectively toggle between the two positions."
              (line-number-at-pos pt))
           (goto-char prompt-pt)
         (beginning-of-line)))))
+(defun vterm-end-of-line ()
+  "Move point to the end of the line, bypassing line wraps."
+  (interactive)
+  (goto-char (vterm--get-end-of-line)))
+
 
 (defun vterm-reset-cursor-point ()
   "Make sure the cursor at the right postion."
--- a/vterm.el
+++ b/vterm.el
@@ -929,12 +929,12 @@ in README."
   "Get the position of the end of current prompt.
 More information see `vterm--prompt-tracking-enabled-p' and
 `Directory tracking and Prompt tracking'in README. "
-  (let ((cur-point (point)))
+  (let ((end-point (end-of-line)))
     (save-excursion
-      (vterm-beginning-of-line)
+      (beginning-of-line)
       (if vterm-copy-use-vterm-prompt
-          (goto-char (next-single-char-property-change (point) 'vterm-prompt nil cur-point))
-      (search-forward-regexp vterm-copy-prompt-regexp cur-point 'noerror))
+          (goto-char (next-single-char-property-change (point) 'vterm-prompt nil end-point))
+      (search-forward-regexp vterm-copy-prompt-regexp end-point 'noerror))
     (point))))
 
 (defun vterm--at-prompt-p ()
@@ -950,7 +950,7 @@ Effectively toggle between the two positions."
   (interactive)
   (if (vterm--at-prompt-p)
       (goto-char (vterm--get-beginning-of-line))
-    (goto-char (vterm--get-prompt-point)))
+    (goto-char (vterm--get-prompt-point))))
 
 (defun vterm-end-of-line ()
   "Move point to the end of the line, bypassing line wraps."
--- a/vterm.el
+++ b/vterm.el
@@ -155,13 +155,8 @@ party to commandeer your editor."
   :type 'boolean
   :group 'vterm)
 
-(defcustom vterm-copy-prompt-regexp term-prompt-regexp
-  "A regexp to find the end of the prompt from the start of the line."
-  :type 'regexp
-  :group 'vterm)
-
-(defcustom vterm-copy-use-vterm-prompt t
-  "Should we use the vterm prompt tracker or the search from `vterm-copy-prompt-regexp'?"
+(defcustom vterm-use-vterm-prompt t
+  "Should we use the vterm prompt tracker or the search from `term-prompt-regexp'?"
   :type 'boolean
   :group 'vterm)
 
@@ -465,7 +460,7 @@ will invert `vterm-copy-exclude-prompt` for that call."
     ;; Are we excluding the prompt?
     (if (or (and vterm-copy-exclude-prompt (not arg))
             (and (not vterm-copy-exclude-prompt) arg))
-        (goto-char (vterm--get-prompt-point)))
+        (goto-char (or (vterm--get-prompt-point)))
     (set-mark (point))
     (goto-char (vterm--get-end-of-line))
     (kill-ring-save (region-beginning) (region-end)))
@@ -886,7 +881,8 @@ in README."
 (defun vterm-next-prompt (n)
   "Move to end of Nth next prompt in the buffer."
   (interactive "p")
-  (if (vterm--prompt-tracking-enabled-p)
+  (if (and vterm-use-vterm-prompt
+           (vterm--prompt-tracking-enabled-p))
       (let ((pt (point)) )
         (goto-char (vterm--get-prompt-point))
         (cl-loop repeat (or n 1) do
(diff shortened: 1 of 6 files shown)


vterm.el
```diff
@@ -432,14 +447,34 @@ This is the value of `next-error-function' in Compilation buffers."
...
+                (and (not vterm-copy-exclude-prompt) arg))
+        ;; Using vterm's prompt detection
+        (if vterm-copy-use-vterm-prompt
+            (vterm-beginning-of-line)
+          ;; Using vterm-copy-prompt-regexp
+          (unless (re-search-forward vterm-copy-prompt-regexp (save-excursion (end-of-line) (point)) 'noerror)
+            ;; Move to the front of the line if no prompt found
+            (beginning-of-line))))
+      (set-mark (point))
+      (end-of-line))
```
jixiuf, line 476: Can you skip the `fake-newline` 
You can check whether the newline is faked by `(get-text-property (point) 'vterm-line-wrap)` when the point at the end of line .
and see `vterm--filter-buffer-substring` and `vterm--remove-fake-newlines`  and https://github.com/akermu/emacs-libvterm/pull/274

Sbozzolo:
Thanks, excellent work! 

I think that we don't need to support finding the prompt via regexp. The preferred way to identify the prompt is by tagging the prompt with `'vterm-prompt`. Removing that part of the code will bring the complexity down and we will have only one additional custom variable.

vterm.el
```diff
@@ -432,14 +447,34 @@ This is the value of `next-error-function' in Compilation buffers."
     (use-local-map vterm-mode-map)
     (vterm-send-start)))
 
-(defun vterm-copy-mode-done ()
-  "Save the active region to the kill ring and exit `vterm-copy-mode'."
-  (interactive)
+(defun vterm-copy-mode-done (arg)
+  "If a region is active then kill it, otherwise kill the current line, then exit `vterm-copy-mode'.
```
Sbozzolo, line 451: It is better if the first sentence of a docstring stays under 67 characters. What you had before is fine (since you explain what the function does after the first sentence).

zallison: Changed back to the first draft.

zallison:
> I think that we don't need to support finding the prompt via regexp. The preferred way to identify the prompt is by tagging the prompt with `'vterm-prompt`.

The `vterm-prompt` isn't available everywhere, as I noticed as I ssh'd to a router (can't change the prompt there) and wanted to copy the command to repeat on another.  Same for attaching to a docker container.

Also I can't get the prompt tag to work *with my prompt*.  Directory tracking works, remote directory tracking works, prompt tag, no. Setting it back to a simple prompt and it works, but I can't sacrifice my prompt!

So there are couple examples of when the `regexp` version would be handy, and that's just from my work today.

> Removing that part of the code will bring the complexity down and we will have only one additional custom variable.

That is true, but it's not a *lot* of complexity. I think the benefits of the added flexibility are worth it.

vterm.el
zallison: That is a very good idea.  I'll have to read about `fake-newlines` and see how to do that.

Sbozzolo:
> Also I can't get the prompt tag to work with my prompt. Directory tracking works, remote directory tracking works, prompt tag, no. Setting it back to a simple prompt and it works, but I can't sacrifice my prompt!

Well, this sounds like a perfect chance to fix some bugs in vterm. ;) 

> That is true, but it's not a lot of complexity. I think the benefits of the added flexibility are worth it.

I don't disagree with this. But, at the moment `vterm` is not supporting prompt detection via regexp elsewhere. I can see in the future using the `'vterm-prompt` property if available, and falling back to the regexp if that is not available. But I think this will require it's own infrastructure. 

If you are interested in implementing this project, I think it should be doable. The goal would be extending the function `vterm--get-prompt-point` so that it uses the text property when it can, and it uses the prompt detection via regexp otherwise. Then, we should make sure that this function is used consistently in the code. In doing this, it would be great if the documentation and docstrings were improved. The cherry on the cake would be writing a couple of tests.

zallison:
I refactored a LOT of existing code and pulled some functionality out into helper functions to keep things cleaner and work with the `fake-newlines`.

The only difference between the vterm flag mode and the regexp mode happens in `vterm--get-prompt-point`.  It's one if statement, with one line for each option.  Everywhere else the code is the same.  So that minimizes code complexity a lot, as did the refactor.

As I said I've been having trouble with the vterm flag, so I'd really appreciate some help to make sure that it's correct.  The regexp version works like a charm so if that line is correct for the vterm flag it should work perfectly there too!

I still like the regexp version as it works locally, on remote routers with unchangeable prompts, when inside a docker container, basically everywhere.  Also it doesn't require any shell side configuration.  Hopefully that's general enough to work with everything else.

I'm also not sure about removing the `fake-newlines`.  Just copying it into the kill ring works and paste works just as expected.

zallison:
I think that about does it.  Let me know how the vterm prompt flag version works, and if we need to do something about the `fake-newlines` `when we copy-to-kill-ring`.  Does it even copy text properties?

jixiuf:
```
Debugger entered--Lisp error: (args-out-of-range 0 0)
  get-text-property(0 vterm-line-wrap)
  (while (get-text-property (1- (point)) 'vterm-line-wrap) (forward-char -1) (beginning-of-line))
  (save-excursion (beginning-of-line) (while (get-text-property (1- (point)) 'vterm-line-wrap) (forward-char -1) (beginning-of-line)) (point))
  vterm--get-beginning-of-line()
  (goto-char (vterm--get-beginning-of-line))
  (if (vterm--at-prompt-p) (goto-char (vterm--get-beginning-of-line)) (goto-char (vterm--get-prompt-point)))
  vterm-beginning-of-line()
  (cond ((derived-mode-p 'term-mode) (term-bol nil)) ((derived-mode-p 'vterm-mode) (vterm-beginning-of-line)) ((derived-mode-p 'eshell-mode) (let ((oldpos (point))) (eshell-bol) (and (= oldpos (point)) (beginning-of-line)))) (t (let ((oldpos (point))) (beginning-of-line) (and (= oldpos (point)) (back-to-indentation)))))
  smart-beginning-of-line()
  funcall-interactively(smart-beginning-of-line)
  call-interactively(smart-beginning-of-line nil nil)
  command-execute(smart-beginning-of-line)
```

jixiuf:
`vterm-beginning-of-line` behave oddly when current line does not contains a  prompt.

zallison:
> `vterm-beginning-of-line` behave oddly when current line does not contains a prompt.

This appears to be `vterm--get-prompt-point` under the vterm-flag mode.  I expanded the `ignore-errors` around the entire `if` function, which should allow for normal operation on lines without a prompt.

At this point I need someone with a working prompt flag to look into `vterm--get-prompt-point`.  I'm doing what I can, but it's hard when I can't test it myself.

You can use `eval-expression` (`M-:`) and then `(goto-char (vterm--get-prompt-point))` to see where it thinks the prompt is.

Once `vterm--get-prompt-point` works correctly everything else should work too.

vterm.el
```diff
@@ -930,9 +930,9 @@ More information see `vterm--prompt-tracking-enabled-p' and
   (let ((end-point (vterm--get-end-of-line)))
     (save-excursion
       (beginning-of-line)
-      (if vterm-copy-use-vterm-prompt
-          (goto-char (next-single-char-property-change (point) 'vterm-prompt nil end-point))
-        (ignore-errors
+      (ignore-errors ;; if no point prompt, don't error out, stay at the beginning-of-line
+        (if vterm-copy-use-vterm-prompt
+            (goto-char (next-single-char-property-change (point) 'vterm-prompt nil end-point))
```
zallison, line 935: This is the line that needs review by someone with working vterm-prompt setup.

jixiuf:
@zallison  There are 18commits now Can you rebase and rewrite  your commits and do a force push.

vterm.el
```diff
@@ -868,59 +893,73 @@ in README."
...
         (cl-loop repeat (or n 1) do
                  (setq pt (previous-single-property-change (1- (point)) 'vterm-prompt))
                  (when pt (goto-char (1- pt)))))
     (term-previous-prompt n)))
 
+(defun vterm--get-beginning-of-line ()
+  "Find the start of the line, bypassing line wraps."
+  (save-excursion
+    (beginning-of-line)
+    (while (and (not (bobp))
```
Sbozzolo, line 909: I haven't fully read and understood the code, but I noticed this while loop that deals with `vterm-line-warp. Is this necessary to look at every single char one at the time? Why not something like `previous-single-property-change`?

jixiuf: line by line

Sbozzolo: Oh, yes. My bad.

zallison:
There we go, squashed down to 2 commits.

zallison:
@jixiuf I've reviewed the changes in your push.  They mostly look good to me, except now I'm unable to copy any line but a prompt line.  Selecting  an output line and running vterm-copy-done gives me the previous command instead of the line I'm on.

zallison:
I merged in @jixiuf's branch and fixed vterm--get-prompt-point for regexp searching.

(Side note, I figured out why the vterm-prompt marking isn't working.  My bash prompt handler runs it before printing the prompt, so the mark is set at the beginning of the line.  So hopefully I can fix that and send it upstream too)

vterm.el
```diff
@@ -927,14 +927,20 @@ in README."
...
-          (search-forward-regexp vterm-copy-prompt-regexp end-point)))
-      (point))))
+      (if (and vterm-use-vterm-prompt
+               (vterm--prompt-tracking-enabled-p))
+          (if (get-text-property end-point 'vterm-prompt)
+              end-point
+            (setq prompt-point (previous-single-property-change end-point 'vterm-prompt))
+            (when prompt-point (setq prompt-point (1- prompt-point))))
+        (goto-char end-point)
+        (if (search-backward-regexp term-prompt-regexp (vterm--get-beginning-of-line) 'noerror)
```
jixiuf, line 940: `vterm--get-prompt-point`:  Get the position of the end of current prompt. 
You should not  set the bound for `search-backward-regexp` .

`(vterm--at-prompt-p)` would return `t` when the point is at the beginning line ,this is not correct.

zallison:
I believe this addresses your concerns and now behaves as expected.

Sbozzolo:
Can we revert this PR and squash the 21 commits into one?

jixiuf:
It is squashed.

Sbozzolo:
Oh, sorry. In the PR page I saw 21 commits, so I thought that it was not reduced to one.

zallison:
Yaay, thanks for you help and the rest of the work on this package.

If there's any other issue you'd like help with let me know and I can take a look.

Sbozzolo:
> If there's any other issue you'd like help with let me know and I can take a look.

Our CI is extremely rudimentary, it would great if we could at least make sure that the package compiles file on all the main platforms.

￼

PR #289, opened 2020-04-11
Remove, vterm-title-functions, add vterm-buffer-name-string

Sbozzolo:
`vterm-title-functions` main purpose is to rename vterm buffers. But, at the moment, it is overly complicated. This PR substitutes `vterm-title-functions` with a new simpler interface that automatically renames buffers with a format string defined by the user. Now, users can simply set `vterm-buffer-name-string` to `%s` to have buffers named after the title (if properly configured shell-side).

If I am missing some use of `vterm-title-functions`, please let me know.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -1103,7 +1103,7 @@ More information see `vterm--prompt-tracking-enabled-p' and
           (vterm--get-beginning-of-line))))))
 
 (defun vterm--at-prompt-p ()
-  "Return t if the cursor postion is at shell prompt."
+  "Return t if the cursor position is at shell prompt."
   (= (point) (vterm--get-prompt-point)))
 
 (defun vterm-beginning-of-line ()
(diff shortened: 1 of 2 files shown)


vterm.el
```diff
@@ -818,8 +820,9 @@ If N is negative backward-line from end of buffer."
     (eq 0 (forward-line n)))))
 
 (defun vterm--set-title (title)
-  "Run the `vterm--set-title-hook' with TITLE as argument."
-  (run-hook-with-args 'vterm-set-title-functions title))
+  "Set the name of the buffer according to `vterm-buffer-name-string'."
+  (when vterm-buffer-name-string
+    (rename-buffer (format vterm-buffer-name-string title))))
```
jixiuf, line 825: suggest change to 
```
 (rename-buffer (format vterm-buffer-name-string title) t)))
```
```
rename-buffer is an interactive built-in function in ‘C source code’.

(rename-buffer NEWNAME &optional UNIQUE)

  Probably introduced at or before Emacs version 1.10.

This function has :around advice: ‘uniquify--rename-buffer-advice’.

Change current buffer’s name to NEWNAME (a string).
If second arg UNIQUE is nil or omitted, it is an error if a
buffer named NEWNAME already exists.
If UNIQUE is non-nil, come up with a new name using
‘generate-new-buffer-name’.
Interactively, you can set UNIQUE with a prefix argument.
We return the name we actually gave the buffer.
This does not change the name of the visited file (if any).
```

PR #290, opened 2020-04-11
Rename vterm-clear-scrollback to vterm-clear-scrollback-when-clearing

Sbozzolo:
The option `vterm-clear-scrollback` has a non-descriptive name that clashed with a function named in the same way.


condy0919:
It also should be `make-obsolete-variable`

Sbozzolo:
Thanks, added.

PR #293, opened 2020-04-12
Automatic detection for system libvterm first

condy0919:
It is mainly for `vterm-module-cmake-args` when it is just installed, it is not available. So I want to use this variable to let `vterm` use the system `libvterm` package. This is done through `CMakeLists.txt`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,9 +35,9 @@ if (USE_SYSTEM_LIBVTERM)
 
   # vterm.h is found.
   if (LIBVTERM_INCLUDE_DIR)
-    message(STATUS "System libvterm detected.")
+    message(STATUS "System libvterm detected")
   else()
-    message(STATUS "System libvterm not found. libvterm will be downloaded and compiled as part of the build process.")
+    message(STATUS "System libvterm not found: libvterm will be downloaded and compiled as part of the build process")
   endif()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,17 +24,24 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
 endif()
 
 # Look for the header file.
-option(USE_SYSTEM_LIBVTERM "Use system libvterm instead of the vendored version." OFF)
+option(USE_SYSTEM_LIBVTERM "Use system libvterm instead of the vendored version." ON)
 
-if(USE_SYSTEM_LIBVTERM)
-  find_path(LIBVTERM_INCLUDE_DIR NAMES
-    vterm.h
+# Try to find the libvterm in system.
+if (USE_SYSTEM_LIBVTERM)
+  # try to find the vterm.h header file.
+  find_path(LIBVTERM_INCLUDE_DIR
+    NAMES vterm.h
     )
 
-  if(NOT LIBVTERM_INCLUDE_DIR)
-    message(FATAL_ERROR "vterm.h not found")
+  # vterm.h is found.
+  if (LIBVTERM_INCLUDE_DIR)
+    message(STATUS "System libvterm detected.")
+  else()
+    message(STATUS "System libvterm not found. libvterm will be downloaded and compiled as part of the build process.")
   endif()
+endif()
 
+if (LIBVTERM_INCLUDE_DIR)
   find_library(LIBVTERM_LIBRARY NAMES
     vterm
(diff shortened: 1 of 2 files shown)


codygman:
This will be great for nix based emacs configs..  maybe it's why I couldn't get vterm working on nix actually.

condy0919:
@Sbozzolo @jixiuf 

It has been pending for several days...

Sbozzolo:
Sorry for the delay. This PR touches the build system, so it has to be checked and tested carefully before being merged. I'll try to get it done in the next days.

CMakeLists.txt
```diff
@@ -23,9 +23,27 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
+# Try to find the libvterm in system.
+if(ENABLE_LIBVTERM_AUTO_DETECTION AND NOT USE_SYSTEM_LIBVTERM)
+  # Try to find the vterm.h header file.
+  find_path(LIBVTERM_INCLUDE_DIR
+    NAMES vterm.h
+    )
+
+  # vterm.h is found
+  if (LIBVTERM_INCLUDE_DIR)
+    set(USE_SYSTEM_LIBVTERM ON CACHE BOOL "vterm.h is found. Use system libvterm directly." FORCE)
```
Sbozzolo, line 41: Could you please add a message to inform the user that we found `libvterm`?
Something like `message(STATUS "System libvterm detected")` would work

CMakeLists.txt
```diff
@@ -23,9 +23,27 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
+  else()
+    message(NOTICE "system libvterm not found.")
```
Sbozzolo, line 43: I would change this from NOTICE to STATUS, capitalize it, and say that we will download libvterm:
```suggestion
    message(STATUS "System libvterm not found. libvterm will be downloaded and compiled as part of the build process.")`.
```

CMakeLists.txt
Sbozzolo, line 41: The wording of the docstring is a little bit inconsistent with the rest of the file. Maybe copy the one few lines above?
(I honestly don't understand why this is needed in the first place, I am not a CMake expert)

condy0919: Sure. It makes sense

condy0919: Bypass the original check statement by setting `USE_SYSTEM_LIBVTERM` to `ON`

condy0919:
- Added documentation in README.

README.md
```diff
@@ -64,12 +64,11 @@ Clone the repository:
...
-system, change `cmake ..` with `cmake -DUSE_SYSTEM_LIBVTERM=yes ..` in the
-following instructions. If `-DUSE_SYSTEM_LIBVTERM` is not explicitly set to
-`yes` (or if it is set to `no`), emacs-libvterm will download the latest version
-available of libvterm (from [here](https://github.com/neovim/libvterm)), compile
-it, and use it.
+By default, system libvterm is automatically detected. If it is not installed,
+emacs-libvterm will download the latest version available of libvterm (from
+[here](https://github.com/neovim/libvterm)), compile it, and use it. If you want
+to use the vendored version, set `ENABLE_LIBVTERM_AUTO_DETECTION` to `Off`. The
+option will be passed into `cmake` through `vterm-module-cmake-args`.
```
Sbozzolo, line 72: I realized that the current setup is more complicated than it should be. 
I think what `vterm` should do is:
- Check if it can find libvterm
- If it can, use it _unless_ `USE_SYSTEM_LIBVTERM` is set to no. In this case, download libvterm
- If it cannot find system-libvterm, download it.

Hence, I suggest we remove the option for `ENABLE_LIBVTERM_AUTO_DETECTION` and change the way `USE_SYSTEM_LIBVTERM` is used.


```suggestion
By default, vterm will try to find if libvterm is installed. If it is 
not found, vterm will download the latest version available of libvterm (from 
[upstream](https://github.com/neovim/libvterm)), compile it, and use it. If you always
want to use this version as opposed to the one on your system, set `USE_SYSTEM_LIBVTERM` 
to `no`. To do this, change `cmake ..` with `cmake -DUSE_SYSTEM_LIBVTERM=no ..` in the
following instructions.
￼```

Sbozzolo:
Perfect, let me test this, and this I'll merge it.

PR #294, opened 2020-04-12
First build of continuous integration for emacs-libvterm

zallison:
This builds on `ubuntu-latest` and `macos-latest`.

Using github actions this:
- retrieves the source
- installs any dependencies (libtool for ubuntu)
- Runs the install as described in README.md
  - makes the Build directory
  - enters it and runs `cmake ..`, returns to home
  - enters it and runs `make`, returns to home

When tests are added we can added `make test` step as well.

I left it as only running on the vterm-ci branch until it's decided how this will fit into the workflow (e.g. does it run on a testing branch or master or both?)

Building on `windows-latest` looks like it should be possible, the image includes cmake support, but I haven't used windows in ages and couldn't figure out how to install libtool.  Someone with more recent experience might have better luck.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,53 @@
+# emacs-libvterm CI
+
+name: vterm-continuous-integration
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the vterm-ci branch.
+on:
+  push:
+    branches: [ vterm-ci ]
+  pull_request:
+    branches: [ vterm-ci ]
+
+jobs:
+  ## Build for ubuntu and macos
+  build:
+    strategy:
+      matrix:
+        # The types of runner that the job will run on
+        os: [ ubuntu-latest, macos-latest ]
+
+    runs-on: ${{ matrix.os }}
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out the repository under $GITHUB_WORKSPACE
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


Sbozzolo:
Thanks. This is a great starting point. Ideally, we would like to cover at least:
1. Ubuntu latest, Ubuntu latest LTS, MacOS
2. Compiling libvterm, and using the system libterm.

We should also run vterm after the compilation (just use `make run`).

zallison:
1) I added `ubuntu-16.04` as there is no explicit image for Ubuntu latest LTS

2) I'll leave the actual testing in the capable hands of the developers.

3) `make run` won't work on github because first there's no GUI available (by default or that I could figure out how to enable, anyone know?) and secondly it doesn't ever exit.  So the CI test would run for 6 hours and then fail.

Sbozzolo:
Thanks, that's a good start. It would be useful to add the case where the compilation is against a system libvterm, as well as the recently released Ubuntu.

zallison:
The most recent Ubuntu is not available in github actions yet.  You can see the supported images here: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners

I think testing it against the system libvterm is a good idea too.  I haven't dealt with C in ages; I wouldn't even know where to start.  I think someone who knows what they're doing would have much better luck.

Sbozzolo:
Thank you very much for your contribution. I took some elements and put them into #410, which has significant improvements. 

Sorry for the embarrassing response time.

I close this in favor of #410.

Issue #298, opened 2020-04-16
The sequence 'M-x vterm M-x vterm-mode M-x vterm' causes emacs to crash with SIGSEGV


iwpiv:
Confirmed to crash with vanilla Emacs, Version: 26.3, fedora 30

Sbozzolo:
Interesting. You can crash (vanilla) Emacs just by opening a new vterm buffer and switching to fundamental mode.

We need to fix this.

Sbozzolo:
#299 will prevent users from changing major-mode, so this problem should be fixed.

PR #299, opened 2020-04-17
Prevent change of major-mode

Sbozzolo:
Addresses #298.

It makes no sense to change major-mode. So we prevent users from doing that. A similar (more drastic) approach is followed by EXWM.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -410,6 +410,12 @@ The units are seconds.")
            :sentinel (when (or vterm-exit-functions
                                vterm-kill-buffer-on-exit)
                        #'vterm--sentinel))))
+
+  ;; Change major-mode is not allowed
+  (add-hook 'change-major-mode-hook
+            (lambda () (interactive)
+               (user-error "You cannot change major mode in vterm buffers")) nil t)
+
   (vterm--set-pty-name vterm--term (process-tty-name vterm--process))
   (process-put vterm--process 'adjust-window-size-function
                #'vterm--window-adjust-process-window-size)


vterm.el
```diff
@@ -410,6 +410,12 @@ The units are seconds.")
            :sentinel (when (or vterm-exit-functions
                                vterm-kill-buffer-on-exit)
                        #'vterm--sentinel))))
+
+  ;; Change major-mode is not allowed
+  (add-hook 'change-major-mode-hook
+            #'(lambda () (interactive)
```
dakra, line 416: Don't think it matters much but you don't have to sharp quote lambda.

vterm.el
```diff
@@ -410,6 +410,12 @@ The units are seconds.")
...
+                (user-error "You cannot change major mode in vterm buffers")) -90 t)
```
fbergroth, line 417: -90? I think you want `nil` for append.

Sbozzolo: I want to prepend. The error has to be thrown immediately, or Emacs will crash.

fbergroth: Any non-nil value (which -90 is) will append. Only `nil` will prepend.

Sbozzolo: Thanks for catching that. I experimented with `nil`, but it seemed that it didn't work. Now it's fixed.

vterm.el
Sbozzolo: Fixed. Thanks.

dakra: Sorry. I meant lambda's are self quoting and you shouldn't quote them at all.
A good read is https://nullprogram.com/blog/2017/12/14/

PR #303, opened 2020-04-22
fix (vterm--get-prompt-point) return nil when no prompt found

### LINKED ISSUE #300: vetrm-beginning-of-line fails in vterm-copy-mode
If I enter to `vterm-copy-mode`, then move point off the prompt (say press <kbd>C-p</kbd>), and then I want to move point to the beginning of the line with <kbd>C-a</kbd>, error occurs:
```
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  =(1 nil)
  vterm--at-prompt-p()
  (if (vterm--at-prompt-p) (goto-char (vterm--get-beginning-of-line)) (goto-char (max (vterm--get-prompt-point) (vterm--get-beginning-of-line))))
  vterm-beginning-of-line()
  funcall-interactively(vterm-beginning-of-line)
  call-interactively(vterm-beginning-of-line nil nil)
  command-execute(vterm-beginning-of-line)
```
If point is at prompt (last terminal line), then no error, point goes to the beginning of the line

Thanks

jixiuf:
this should fix #300

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -600,7 +600,7 @@ will invert `vterm-copy-exclude-prompt' for that call."
     ;; Are we excluding the prompt?
     (if (or (and vterm-copy-exclude-prompt (not arg))
             (and (not vterm-copy-exclude-prompt) arg))
-        (goto-char (max (vterm--get-prompt-point)
+        (goto-char (max (or (vterm--get-prompt-point) 0)
                         (vterm--get-beginning-of-line))))
     (set-mark (point))
     (goto-char (vterm--get-end-of-line)))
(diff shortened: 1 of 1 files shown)


vterm.el
```diff
@@ -1103,7 +1106,7 @@ More information see `vterm--prompt-tracking-enabled-p' and
 
 (defun vterm--at-prompt-p ()
   "Return t if the cursor postion is at shell prompt."
-  (= (point) (vterm--get-prompt-point)))
+  (= (point) (or (vterm--get-prompt-point) 0)))
```
Sbozzolo, line 1109: I think the name is misleading (ie, it doesn't convey what the function does): `vterm--at-prompt-p` should return t only when the point is at the prompt, if `vterm--get-prompt-point` returns nil, it means that we cannot locate the prompt, so this function should fail.

Also, we should document that if the prompt detection fails, we use the beginning of the line instead.

jixiuf: when tail -f  
and all promp  disappeard  it return nil

Sbozzolo: `(or nil 0)` returns 0, so I don't see how this function can return nil

jixiuf: when `(vterm--get-prompt-point)` is a number , `(or  (vterm--get-prompt-point) 0) `would return that number.
when  `(vterm--get-prompt-point)` is nil   `(or  (vterm--get-prompt-point) 0)`  would return 0 

`(= (point) 0)` would return nil, because `(point)` always >0

jixiuf:
the force-push just rebase origin/master

vterm.el
Sbozzolo: Gotcha! I completely overlooked the function. Thanks for the detailed explanation.

Then, it looks good to me.

PR #306, opened 2020-04-25
Update vterm-use-vterm-prompt docstring

PuercoPop:
Spell out which prompt tracking strategy is used for each value

Thanks for libvterm!

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -259,9 +259,9 @@ rendered as normal ones."
   :group 'vterm)
 
 (defcustom vterm-use-vterm-prompt t
-  "If t track the prompt using the vterm prompt tracker. If set
-to nil use the value of `term-prompt-regexp' to track the
-prompt."
+  "If not nil track the prompt using the prompt tracker
+functionality provided by vterm. If set to nil use the value of
+`term-prompt-regexp' to track the prompt."
   :type 'boolean
   :group 'vterm)
 
--- a/vterm.el
+++ b/vterm.el
@@ -259,7 +259,9 @@ rendered as normal ones."
   :group 'vterm)
 
 (defcustom vterm-use-vterm-prompt t
-  "Should we use the vterm prompt tracker or the search from `term-prompt-regexp'?"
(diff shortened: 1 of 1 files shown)


Sbozzolo:
Thanks for your contribution. A couple of things:
1. In Emacs lisp any symbol that is not `nil` has value true. Therefore, instead of saying "if t", it is better to say "if not nil" (as according to official guidelines)
2. After periods there should be two spaces, not one.
3. Typically, docstrings start with a short description, have an empty line, and additional information after that.

In the spirit of improving the docstring you touched, I think it would be better to add a more in-depth description of what is the "vterm prompt tracker", when it should be used, and when not. If you can update your PR with this, that would be great.

PuercoPop:
> 1. In Emacs lisp any symbol that is not nil has value true. Therefore, instead of saying "if t", it is better to say "if not nil" (as according to official guidelines)

Done. However I'll note that other defcustom docstrings in this package use 'if set to t' phrasing.

> 2. After periods there should be two spaces, not one.

Done.

> 3. Typically, docstrings start with a short description, have an empty line, and additional information after that.
> 
> In the spirit of improving the docstring you touched, I think it would be better to add a more in-depth description of what is the "vterm prompt tracker", when it should be used, and when not. If

I agree with that, however I don't think I am able to do a proper description of the custom option given that I've just started using this package and I'm not too familiar with its features yet. For example it appears that if the shell-side configuration is missing, even if `vterm-use-vterm-prompt` is set to `t`, `term-prompt-regex` is used as a fallback strategy?

The problem I address with the current docstring is that it is answering a 'Do you want A or B?' question with 'yes or no'. Given that yes or no is not an intelligible answer, after reading the docstring, the user is left wondering what option maps to which value.

Sbozzolo:
I am sorry for the extremely long response time. I update and expand the commentary and docstrings in #364

PuercoPop:
> I am sorry for the extremely long response time.

No need to apologize for how you choose to spend your free time. Thanks for updating the docs :tada:. Closing this PR as it is subsumed by #364 !

PR #318, opened 2020-05-06
Add shell configs

arjanadriaanse:
Combining the described configurations into separate files makes it easier for users to configure their shell. Also, package managers could automatically install them system-wide, see for example [a Gentoo ebuild](https://github.com/gentoo/guru/blob/master/app-emacs/vterm/vterm-20200504.544.ebuild).

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,15 @@ if [[ "$INSIDE_EMACS" = 'vterm' ]]; then
     }
 fi
 ```
+For `fish`:
+```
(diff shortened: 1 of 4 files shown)


Sbozzolo:
Thanks, @arjanadriaanse. I think this is a good idea, but we need to have documentation inside the files. Users should be able to understand what every part of the code is for. 

Also, could you please make squash the commits to only one?

Thanks.

PR #320, opened 2020-05-09
Dirty hack for local directory track

condy0919:


condy0919:
@Sbozzolo @jixiuf

jixiuf:
Currently  `Directory tracking and Prompt tracking`  are using one  escaped sequences `51;A`
Maybe we can add another escaped sequences  in case someone only want to enable one of them.
we can change 
```
vterm_prompt_end() {
    vterm_printf "51;A$(whoami)@$(hostname):$(pwd)";
}
PROMPT=$PROMPT'%{$(vterm_prompt_end)%}'

```
to 
```
vterm_prompt_end() {
    vterm_printf "51;A";
}
PROMPT=$PROMPT'%{$(vterm_prompt_end)%}'
###############################
add-zsh-hook -Uz chpwd (){
            vterm_printf "51;B$(whoami)@$(hostname):$(pwd)"
}
```
@Sbozzolo What do you think.

Sbozzolo:
> Currently `Directory tracking and Prompt tracking` are using one escaped sequences `51;A`
> Maybe we can add another escaped sequences in case someone only want to enable one of them.
> we can change
> 
> ```
> vterm_prompt_end() {
>     vterm_printf "51;A$(whoami)@$(hostname):$(pwd)";
> }
> PROMPT=$PROMPT'%{$(vterm_prompt_end)%}'
> ```
> 
> to
> 
> ```
> vterm_prompt_end() {
>     vterm_printf "51;A";
> }
> PROMPT=$PROMPT'%{$(vterm_prompt_end)%}'
> ###############################
> add-zsh-hook -Uz chpwd (){
>             vterm_printf "51;B$(whoami)@$(hostname):$(pwd)"
> }
> ```
> 
> @Sbozzolo What do you think.

Can you give me a use case? If one wants to have only the directory, they can already do it with the current infrastructure.

@condy0919, thanks for this contribution. Why/how do you use this? One problem I see is that the hack fail on remote machines, where I'd think it would be the most useful to have something that doesn't require the customization of the prompt.

condy0919:
Yes, it works only on local machines. Document it in `README` to cheer up those who don't want to change the local shell configuration

condy0919:
https://www.reddit.com/r/emacs/comments/gp7s4i/directory_tracking_in_ansiterm_with_without/

The reddit guy want local directory track without shell side configuration

condy0919:
Close for #335

PR #335, opened 2020-05-27
Fix message passage for long commands, add more FAQs

Sbozzolo:
@condy0919 I added your PR #320 to the FAQ, what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,8 +555,8 @@ On most GNU/Linux systems, you can read current directory from `/proc`:
   (interactive)
   (when vterm--process
     (let* ((pid (process-id vterm--process))
-           (dir (file-truename (format "/proc/%d/cwd" pid))))
-      (setq-local default-directory dir))))
+           (dir (file-truename (format "/proc/%d/cwd/" pid))))
+      (setq default-directory dir))))
 ```
 A possible application of this function is in combination with `find-file`:
 ```emacs-lisp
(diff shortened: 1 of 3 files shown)


condy0919:
Fine with me

README.md
```diff
@@ -594,6 +543,28 @@ the correct function to yank in vterm buffers.
...
+tracking. If you cannot do it, a possible workaround is the following.
+
+On most GNU/Linux systems, you can read current directory from `/proc`:
+```emacs-lisp
+(defun vterm-directory-sync ()
+  "Synchronize current working directory."
+  (interactive)
+  (when vterm--process
+    (let* ((pid (process-id vterm--process))
+           (dir (file-truename (format "/proc/%d/cwd" pid))))
```
condy0919, line 558: I found `(dir (file-truename (format "/proc/%d/cwd/" pid))))` is better

Sbozzolo: That's what we have here. Am I missing something?

condy0919: `/proc/%d/cwd/` with an additional `/` makes it prettier in `counsel`.

no trialing slash

prettier

README.md
```diff
@@ -594,6 +543,28 @@ the correct function to yank in vterm buffers.
...
+      (setq-local default-directory dir))))
```
brotzeit, line 559: I think `default-directory` becomes automatically buffer local.

Sbozzolo: Thanks

README.md
Sbozzolo: Thanks for the clarification

Issue #352, opened 2020-06-23
How to set unlimited scrollback ?


grizzlypolar:
Currently, one of the commands produces 2K lines to stdout but I have observed that the maximum number of lines in an `vterm` buffer is `1055`. How do I set this to unlimited lines (or at least increase this limit) so that I can go back more in my command output.

grizzlypolar:
I was able to set 
```
(setq vterm-max-scrollback 100000)
```
 and was able to increase the scroll back to `100000`.  I saw the comment here that we can set the upper limit to more than `100000` by setting `SB_MAX` and re-compiling it. 

I'm leaving the issue open because the ideal behaviour here is to have `SB_MAX` to the maximum allowed value on the machine and make the user configure it via `vterm-max-scrollback` in their emacs config.

grizzlypolar:
☝️ Fixes this. 

Feel to accept or reject it before closing this ticket if you have any strong opinions.

@akermu

PR #353, opened 2020-06-23
Setting SB_MAX to the maximum allowed on the machine

grizzlypolar:
Fix for #352 

Setting the scrollback to maximum allowed on the machine.

### DIFF
--- a/vterm-module.h
+++ b/vterm-module.h
@@ -4,11 +4,12 @@
 #include "emacs-module.h"
 #include <inttypes.h>
 #include <stdbool.h>
+#include <limits.h>
 #include <vterm.h>
 
 int plugin_is_GPL_compatible;
 
-#define SB_MAX 100000 // Maximum 'scrollback' value.
+#define SB_MAX INT_MAX // Maximum 'scrollback' value.
 
 #ifndef MIN
 #define MIN(X, Y) ((X) < (Y) ? (X) : (Y))


Sbozzolo:
The reason there is a limit on SB_MAX is (I think) to prevent `vterm` from taking over the entire memory of the system. With a large SB_MAX, the memory footprint of `vterm` could be significant. Imagine, for example, that you are developing a large piece of software and the compilation produces thousands of lines every time. If you have "infinite" scrollback, this would mean keeping in memory thousands of lines that you probably don't need. 

What I can do is to document better how to increase the size of the scroll back in the README for users that are aware of this and know what they are doing.

grizzlypolar:
I may be wrong, but here is my understanding. `SB_MAX` is just the upper bound of the memory consumption. Setting it to a very high value, will allow users to keep so many lines in memory. We already have a safeguard check by having 
```
(defcustom vterm-max-scrollback 1000)
```
which will ensure that not more than 1000 lines in memory. If somebody wants to set this to a high value, it's under the assumption they they have a machine with so much memory capacity. i.e we can keep the defaults very conservative (1000 lines) and if somebody wants to customize / hack it, they are free to do so in their emacs config rather than recompiling vterm.  

Most other terminals(gnome terminal, iterm on mac) allow "infinite" scrollback, so on a decent enough machine (with ~16 Gigs RAM), I can't imagine how it would hog all the memory. Back of envelope calculation of ~200 characters per line and 1 byte per character (not considering UTF-8 for now) shows that each line just needs 200 bytes. i.e
1 line       ->    200 bytes
5 lines     ->    1KB
5k lines   ->    1MB
5M lines  ->    1GB
Even if the user sets `(defcustom vterm-max-scrollback 1000)` to 5 million, the max that terminal can use 1GB.

fbergroth:
Another way would be to add a CMake option for this, if enough users want to change SB_MAX.

Sbozzolo:
The way gnome-terminal implements infinite scrolling is by writing a compressed file to disk. I assume that they are doing this to control memory consumption. Neovim, which has a terminal based on libvterm, is also following the same approach we have right now, with the same exact values for SB_MAX and the default value. I feel that if we want to have extremely large/infinite scrollback, we should do something more sophisticated than just increasing SB_MAX. Of course, nothing prevents users that know what they are doing to change SB_MAX.

milanst:
Isn't the scrollback data already in the emacs buffer? Why is there a need for additional storage?

Issue #363, opened 2020-07-12
Cannot find correct build directory in native-comp emacs


blahgeek:
I'm using `native-comp` branch of emacs (https://akrl.sdf.org/gccemacs.html) which offers greater elisp performance. It will compile all *.el files to *.eln files to a seperate directory.

Since `vterm-module-compile` will locate the source directory using `(locate-library "vterm")`, it cannot find the correct directory in native-comp emacs. Normally the return value is something like `~/.emacs.d/elpa/vterm-20200707.1655`, but with native-comp emacs, it's like `~/.emacs.d/elpa/vterm-20200707.1655/eln-x86_64-pc-linux-gnu-6d32bf10d46bb4cc`.

My current workaround is manually compiling in the correct directory.

mnewt:
PR https://github.com/akermu/emacs-libvterm/pull/373 resolves the issue for me. Can anyone else confirm?

blahgeek:
I can confirm #373 fixes this, thanks for you work

Sbozzolo:
Fixed with #373

Issue #351, opened 2020-06-22
Vterm module must be re-compiled every restart when using straight.el.


akoen:
Whenever Emacs is restarted, the vterm module must be recompiled:

![image](https://user-images.githubusercontent.com/43913902/85240313-4904d880-b427-11ea-8cd9-d6d92f4ed290.png)
Once the module is compiled vterm works normally.

Here is the output of the module build:

```
-- Configuring done
-- Generating done
-- Build files have been written to: /home/alex/.config/emacs/.local/straight/build/vterm/build
Scanning dependencies of target libvterm
[  8%] Creating directories for 'libvterm'
[ 16%] Performing download step (git clone) for 'libvterm'
Cloning into 'libvterm'...
Note: switching to '65dbda3ed214f036ee799d18b2e693a833a0e591'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 65dbda3 Bump version to 0.1.3
[ 25%] No patch step for 'libvterm'
[ 33%] Performing update step for 'libvterm'
[ 41%] No configure step for 'libvterm'
[ 50%] Performing build step for 'libvterm'
TBL src/encoding/DECdrawing.tbl
TBL src/encoding/uk.tbl
CC src/encoding.c
CC src/keyboard.c
CC src/mouse.c
CC src/parser.c
CC src/pen.c
CC src/screen.c
CC src/state.c
CC src/unicode.c
src/unicode.c:306:12: warning: ‘mk_wcswidth_cjk’ defined but not used [-Wunused-function]
  306 | static int mk_wcswidth_cjk(const uint32_t *pwcs, size_t n)
      |            ^~~~~~~~~~~~~~~
src/unicode.c:215:12: warning: ‘mk_wcswidth’ defined but not used [-Wunused-function]
  215 | static int mk_wcswidth(const uint32_t *pwcs, size_t n)
      |            ^~~~~~~~~~~
CC src/vterm.c
LINK libvterm.la
CC bin/unterm.c
CC bin/vterm-ctrl.c
CC bin/vterm-dump.c
[ 58%] No install step for 'libvterm'
[...]

PR #373, opened 2020-07-18
With the native-compile feature enabled, properly find build dir

mnewt:
When the native-compile feature is enabled, `locate-library` returns the path to
the compiled (`.eln`) file rather than the `.el` or `.elc` that would otherwise
be expected. Then `vterm-module-compile` runs the compilation from that
subdirectory and can't find the expected files.

`find-library-name` always returns the `.el` file location.

Example error:
```
CMake Error: The source directory "/Users/user/.emacs.d/straight/build/vterm/eln-x86_64-apple-darwin19.5.0-89a730c11dec204e" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
make: *** No targets specified and no makefile found.  Stop.
```

After this change the compilation files are found as expected.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -104,7 +104,7 @@ the executable."
   (when (vterm-module--cmake-is-available)
     (let* ((vterm-directory
             (shell-quote-argument
-             (file-name-directory (locate-library "vterm"))))
+             (file-name-directory (find-library-name "vterm"))))
            (make-commands
             (concat
              "cd " vterm-directory "; \
--- a/vterm.el
+++ b/vterm.el
@@ -104,6 +104,11 @@ the executable."
   (when (vterm-module--cmake-is-available)
     (let* ((vterm-directory
             (shell-quote-argument
+             ;; NOTE: This is a workaround to fix an issue with how the Emacs
+             ;; feature/native-comp branch changes the result of
+             ;; `(locate-library "vterm")'. See emacs-devel thread
+             ;; https://lists.gnu.org/archive/html/emacs-devel/2020-07/msg00306.html
+             ;; for a discussion.
              (file-name-directory (locate-library "vterm.el" t))))
            (make-commands
             (concat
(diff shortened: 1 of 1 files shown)


blahgeek:
I can confirm this fixes #363

Sbozzolo:
There's an emacs-devel thread about this problem: https://lists.gnu.org/archive/html/emacs-devel/2020-07/msg00306.html

It's not clear to me what approach will be taken by the developers/maintainers and maybe the problem will be fixed upstream. I am not opposed to adding a workaround, though. In the thread they suggest using `(locate-library "vterm.el" t)`. Can you try this?

mnewt:
@Sbozzolo Thanks for testing and sharing that thread. 

I agree continuing to use `locate-library` is desirable, since `find-library-name` is in a package that would otherwise not be loaded by `vterm`. And since it was recommended on the mailing list that carries some weight. 

It's worth noting that they consider it a workaround and hopefully they can figure out a solution that reinstates the original result of `(locate-library "vterm")`. Because of that I added some comments to remind us why it is the way it is and to revisit if appropriate.

Even though this is a workaround, is there a downside to going ahead and implementing this? Maybe if a few more people test we will have some confidence that it won't break anything.

Sbozzolo:
I think this is a good solution. I want to take the chance to see how vterm interacts with straight because there are reports of problems and this is the code I have to look at.

mnewt:
I'm on macOS using straight and it works for me. Result of `(locate-library "vterm.el" t)`:

```
"/Users/mnewt/.emacs.d/straight/build/vterm/vterm.el"
```

I do still see this bug though: https://github.com/akermu/emacs-libvterm/issues/351. I often but not always have to recompile vterm-module in a new Emacs session. But my guess is these are two unrelated issues.

Sbozzolo:
Thanks, looks good to me.

Sbozzolo:
I meant to ask you to squash your commits, but I forgot. I ended up force-pushing it, but now the fix is in.

Thanks!

mnewt:
@Sbozzolo Awesome, thanks! Yeah, definitely the commits make no sense on their own, only the final one is (hopefully) correct). Cheers.

PR #377, opened 2020-07-21
add possibility to store history in file

brotzeit:
This change allows users to track the vterm history in a file. If you intend to merge this feature I will make this right. I just don't want to spend time on it if you don't want this feature to be merged.

I would also add docs on how to add a helm source.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -708,13 +708,35 @@ will invert `vterm-copy-exclude-prompt' for that call."
   (interactive)
   (vterm-send-key "<stop>"))
 
+(defvar vterm-history-save t)
+(defvar vterm-history-file "~/.emacs.d/vterm_history")
+
 (defun vterm-send-return ()
   "Send `C-m' to the libvterm."
   (interactive)
   (when vterm--term
+    (when (and vterm-history-save (not process-running-child-p))
+      (let* ((beg (vterm--get-prompt-point))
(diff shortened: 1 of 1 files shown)


Sbozzolo:
Can you give me some use cases? Why not just use the shell history file? 

The only case I can think of in which the shell file wouldn't suffice is with remote shells.

brotzeit:
Yeah it's mostly about remote shells(I realized that I have to remove process-running-child-p). This is very helpful if you are working a lot with ssh like I do.

Sbozzolo:
At the moment, I would like to focus on "infrastructure" work: bug fixes, documentation, refactoring, and so on. I believe that right now the priority is to make the code more stable and maintainable.

A while ago, I started a personal repo to add some functions to vterm: https://github.com/Sbozzolo/vterm-extra.
At the moment there are only two functions I wrote: one to quickly jump to another open vterm buffer, and the other to write more complex shell commands in a separate buffer. 

I can publish this to MELPA and make it the official "contrib" repo for vterm. You could add your functions there. 
We can upstream functions from vterm-extra to the main repo, once the infrastructure work is done.

What do you think?

brotzeit:
Sorry for the late reply. However I'm not sure if it's important to put optional features into an additional repo. But I will try to add the feature and open a PR in vterm-extra.

Issue #316, opened 2020-05-05
segmentation fault with compilation-shell-minor-mode


tpapp:
I am experimenting with `vterm` in julia-repl (https://github.com/tpapp/julia-repl/pull/84). It works fine, the only issue is that when I click on some file locations in the output to navigate using `compilation-shell-minor-mode`, I *occasionally* get a segfault. But it has proven to be near impossible to reproduce on purpose --- usually it works fine.

The backtrace is
```
emacs[0x512b1e]                                                                                     
emacs[0x4f84da]                                                                                     
emacs[0x510fee]                                                                                     
emacs[0x511218]                                                                                     
emacs[0x511299]                                                                                     
/lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0)[0x7f2d667603c0]                                     
/home/tamas/.emacs.d/elpa/vterm-20200501.2231/vterm-module.so(Fvterm_get_pwd+0x97)[0x7f2d5b48d107]
emacs[0x59c48d]                                                                                     
emacs[0x56f57b]                                                                                     
emacs[0x5ab340]                                                                                     
emacs[0x56f57b]                                                                                     
emacs[0x5ab340]                                                                                     
emacs[0x56f57b]                                                                                     
emacs[0x5ab340]                                                                                     
emacs[0x56f57b]                                                                                     
emacs[0x5ab340]                                                                                     
emacs[0x56f57b]                                                                                     
emacs[0x56bfcd]                                                                                     
emacs[0x56f5fc]                                                                                     
emacs[0x57132a]                                                                                     
emacs[0x56c538]                                                                                     
emacs[0x56f5fc]                                                                                     
emacs[0x5ab340]                                                                                     
emacs[0x56f57b]                                                                                     
emacs[0x56f6ba]                                                                                     
emacs[0x507800]                                                                                     
emacs[0x56e74e]                                                                                     
emacs[0x4f88b4]                                                                                     
emacs[0x56e6bd]                                                                                     
emacs[0x4f884b]                                                                                     
emacs[0x4fd9a3]                                                                                     
emacs[0x4fdce8]                                                                                     
emacs[0x41b991]                                                                                     
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7f2d6644e0b3]
emacs[0x41c69a]                                                                                     
```
[...]

PR #378, opened 2020-07-28
Add a workaround until #316 is fixed.

tpapp:
This is a temporary workaround until #316 can be fixed: it just normalizes invalid rows, preventing a segfault.


Sbozzolo:
I haven't investigated deeply yet, but I think negative rows should be allowed: I think that they are the rows not visible on the screen, but still in the scrollback. 

The code in question is directly from pangoterm, which is an implementation of a terminal emulator from the author of libvterm.

PR #383, opened 2020-08-13
Define key bindings inside defvar forms

tarsius:
The default key bindings should be defined inside the `defvar` form
that defines the respective keymap.

When defining the key bindings separately then reloading the defining
library (e.g. using `eval-buffer`) would discard user customizations
unintentionally, while _intentionally_ discarding all customizations
using `eval-defun` on the `defvar` form becomes impossible.

The `defvar` form has to come before the respective mode is defined,
otherwise `define-derived-mode` would define the keymap, and the
subsequent `defvar` form would fail to set the default value.

This also includes the commit from my previous pull-request, #382,
I can change that if desired.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -438,6 +438,112 @@ Improves performance when receiving large bursts of data.
 If nil, never delay.
 The units are seconds.")
 
+;;; Keybindings
+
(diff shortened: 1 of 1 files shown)


tarsius:
I've fixed that now.

Sbozzolo:
Thanks!

PR #385, opened 2020-08-17
Add implementation of vterm_cmd helper for fish

jamesnvc:


Sbozzolo:
Thanks, can you please add in the docstring that commands have to be defined `vterm-eval-cmds`?
Then squash the commits and I will merge this.

PR #390, opened 2020-08-29
Fix memory leak on finalize, free term->lines

blahgeek:


Sbozzolo:
Thanks! Sorry for the long delay with this one-liner.

PR #392, opened 2020-08-30
add vterm_cmd examples for fish shell to README.md

emacsomancer:
- add vterm_cmd examples for fish shell to README.md
- also, change following line to make the transition clearer


Sbozzolo:
Thanks!

PR #395, opened 2020-09-07
Only modify the prompt if the shell is in vterm

fapdash:
Without this check the config messed up the prompt in the GNOME terminal.

### DIFF
--- a/etc/emacs-vterm-bash.sh
+++ b/etc/emacs-vterm-bash.sh
@@ -53,4 +53,6 @@ PROMPT_COMMAND='echo -ne "\033]0;\h:\w\007"'
 vterm_prompt_end(){
     vterm_printf "51;A$(whoami)@$(hostname):$(pwd)"
 }
-PS1=$PS1'\[$(vterm_prompt_end)\]'
+if [[ "$INSIDE_EMACS" = 'vterm' ]]; then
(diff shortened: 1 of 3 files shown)


Sbozzolo:
We have to find a way to propagate INSIDE_EMACS to ssh sessions, otherwise this will not work.

fapdash:
Can you please explain the problem a bit more? Once we're connected with a remote server via ssh then this terminal session runs with the config of the remote server anways, no?

Sbozzolo:
Yes, if I have these configs on my remote servers, they will not work if INSIDE_EMACS is not defined. So, if we want to modify the prompt (and add the other functions) only in vterm. We have to somehow make sure that INSIDE_EMACS is set to something also when we are ssh-ed somewhere.

fapdash:
Would this do the trick? https://superuser.com/a/702751

Sbozzolo:
Unfortunately, that solution relies on server-side configurations that often cannot be controlled by the user. A workaround would be to pass the information inside an `LC_*` variable, but this is not very satisfying. Some servers block these too. The only one that always go through is `TERM`, but changing it may lead to other problems.

Thaodan:
I think width-editor does this by inserting commands into the session. Another way would be start the shell on the remote with INSIDE_EMACS set.

PR #396, opened 2020-09-09
Fix unsettable vterm-keymap-exceptions

wyuenho:
`vterm-keymap-exceptions` is unsettable due to the `:set` function is calling `vterm--exclude-keys` with a wrong number of arguments, and that it is using `set` instead of `set-default` so the values cannot be persisted. This PR fixes this issue, simplifies the implementation and did some code clean up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,16 @@ behavior can be achieved by using the universal prefix (ie, calling `C-u C-l`).
 
 Shell to run in a new vterm. It defaults to `$SHELL`.
 
+## `vterm-keymap-exceptions`
+
+All the keybindings `C-*`, `M-*` are sent to vterm (where `*` is any character),
+except those defined in the list `vterm-keymap-exceptions`. Some of the ones
+excluded by default are `C-x` or `M-x`, which are captured by Emacs instead.
+
+If you customize this variable, the change will be applied to all the new
+`vterm` sessions, so you may need to restart the ones that are currently
+running.
+
 ## `vterm-term-environment-variable`
 
 Value for the `TERM` environment variable. It defaults to `xterm-256color`. If
@@ -375,7 +385,8 @@ If you want a key to be sent to the terminal, bind it to `vterm--self-insert`,
 or remove it from `vterm-mode-map`. By default, `vterm.el` binds most of the
 `C-<char>` and `M-<char>` keys, `<f1>` through `<f12>` and some special keys
 like `<backspace>` and `<return>`. Sending a keyboard interrupt is bound to `C-c
-C-c`.
+C-c`.  You can customize the default keybindings to ignore by customizing
+`vterm-keymap-exceptions`.
 
 ## Fonts
 
--- a/README.md
+++ b/README.md
@@ -298,13 +298,18 @@ Shell to run in a new vterm. It defaults to `$SHELL`.
 
 ## `vterm-keymap-exceptions`
 
-All the keybindings `C-*`, `M-*` are sent to vterm (where `*` is any character),
-except those defined in the list `vterm-keymap-exceptions`. Some of the ones
-excluded by default are `C-x` or `M-x`, which are captured by Emacs instead.
+All the key bindings `C-*`, `M-*` are sent to vterm (where `*` is any
+character), except those defined in the list `vterm-keymap-exceptions`. Some of
+the ones excluded by default are `C-x` or `M-x`, which are captured by Emacs
+instead.
 
 If you customize this variable, the change will be applied to all the new
 `vterm` sessions, so you may need to restart the ones that are currently
 running.
+
+Please note that if you bind keys to `vterm-mode-map` before a `vterm` session
+is launched, your key bindings overrides the same bindings in
+`vterm-keymap-exceptions`.
 
 ## `vterm-term-environment-variable`
 
@@ -381,12 +386,12 @@ avoid this question and always compile the module, set
 
 ## Keybindings
 
-If you want a key to be sent to the terminal, bind it to `vterm--self-insert`,
-or remove it from `vterm-mode-map`. By default, `vterm.el` binds most of the
-`C-<char>` and `M-<char>` keys, `<f1>` through `<f12>` and some special keys
-like `<backspace>` and `<return>`. Sending a keyboard interrupt is bound to `C-c
-C-c`.  You can customize the default keybindings to ignore by customizing
-`vterm-keymap-exceptions`.
+If you want a key to be sent to the terminal, bind it to `vterm--self-insert`, ,
+remove it from or rebind it in `vterm-mode-map`. By default, `vterm.el` binds
+most of the `C-<char>` and `M-<char>` keys, `<f1>` through `<f12>` and some
+special keys like `<backspace>` and `<return>`. Sending a keyboard interrupt is
+bound to `C-c C-c`.  You can customize the default keybindings to ignore by
+customizing `vterm-keymap-exceptions`.
 
 ## Fonts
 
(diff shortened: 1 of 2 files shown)


wyuenho:
@Sbozzolo mind taking a look at this?

Sbozzolo:
Thanks for your contribution. 

I am sorry I am a little slow at processing issues and PR as of lately (due to my work). 

I will try to have this merged before or during the weekend.

Thanks for your patience.

Sbozzolo:
What is the problem with fixing the setter?
You removed it completely, so if a user customizes `vterm-keymap-exceptions`, they will have to reload `vterm.el` (ie, restart emacs), to see the effects of the change. 

(Thanks for the code cleanup!)

wyuenho:
//  they will have to reload vterm.el (ie, restart emacs), to see the effects of the change.

No you only have to turn off and on `vterm-mode`. This is very much necessary to avoid the mutual dependency in the definition of `vterm-mode-map` and `vterm-keymap-exceptions`. Conceivably we can still dynamically set the mode map in a set function, but that introduces other complexities such as having to diff the set of bindings to remove from the map and the set to add. The existing code just will not do regardless. Let's just keep it simple.

Sbozzolo:
> No you only have to turn off and on vterm-mode.

Ah yes, you are right. I also see what you say about having to diff the old map and the new one.

Okay, I like your approach. Once I have the time to go through the code in more detail, I will merge the PR.

Sbozzolo:
Looks good to me. :)

Two comments to improve your contribution (and to go beyond that!):
1. Can you please add a small section in the README to document this customization? 
2. Can you add a comment in the code to briefly describe how we bind keys and how we exclude the exceptions?
    While the code is clear, the different pieces are scattered around the file, so it would nice to have somewhere 
    (maybe around the definition of `vterm-keymap-exceptions`) a general overview.

Thank you very much!

wyuenho:
Done

Sbozzolo:
I force-pushed to your branch with more comprehensive docs.

However, I noticed a problem. If one assigns one of the keys in `vterm-keymap-exception` to a personal command, the code will cancel that customization (the cl-loop inside the define derived-mode will reset the binding to nil).

You can try with this: `(define-key vterm-mode-map (kbd "C-l") #'(lambda () (interactive) (message "HI")))`

wyuenho:
If you define a personal keybinding and tell vterm to ignore it at the same time, one must win out, they can't be simultaneously true. Either don't define that personal key binding or don't ignore it. It's very simple.

Sbozzolo:
The way I think about `vterm-keymap-exceptions` is "keys that are not to be sent to the terminal". With this mindset, it makes sense to have it among the exceptions, but also assign it to a command. And it is not that `vterm-keymap-exceptions` are plainly ignored. For example, `M-y` is in the exceptions, but it is also bound to a `vterm` command.

I think we should allow users to customize keys in `vterm-keymap-exceptions`.

wyuenho:
> The way I think about vterm-keymap-exceptions is "keys that are not to be sent to the terminal". With this mindset, it makes sense to have it among the exceptions, but also assign it to a command.

That's not even how it works on master now?

wyuenho:
I guarantee you, if you do that, someone eventually is going to come in and ask you why a keybinding is not ignore even when set in the exceptions.

Sbozzolo:
> I guarantee you, if you do that, someone eventually is going to come in and ask you why a keybinding is not ignore even when set in the exceptions.

What I am suggesting is that a key is in the exceptions, it can still be bound to something else, if the user wants.
That is, I think that the following behavior is the one that makes the most sense:
1. If X is in the exceptions, and the user doesn't bind X, then X is bound to nil
2. If X is in the exceptions, and the user binds X, then X is bound to what the user wants
3. If X is not in the exceptions, and the user binds X, then X is bound to what the user wants
4. If X is not in the exceptions, and the user does not bind X, then X is sent to the terminal

If a user binds a key explicitly, we should not discard that customization. That's the spirit of the very first [Emacs coding convention](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html).

We will add documentation to explain this. Would changing the name of the variable help to clarify it?

wyuenho:
How exactly do you plan on distinguish keys bound by the user and keys bound by vterm? Do you want to put a property in every definition of `vterm-define-key` or something?

> If a user binds a key explicitly, we should not discard that customization. That's the spirit of the very first Emacs coding convention.

That document says no such thing, and we are not unbinding a user's keys, we are just following their illogical instructions to bind and ignore at the same time.

I would strongly urge you to not let feature that no one asks for creep into a bug fix.

wyuenho:
Is this going to merged? If not I'll just close it.

Sbozzolo:
It will not be merged in this specific form, but some small variation of it will be merged. The minimum variation could just be clearly documenting the intended behavior.

I wanted to stop and think about this a little bit more, but I haven't found the time yet. Sorry about that.

Please, don't close it.

wyuenho:
You've already documented it perfectly.

Sbozzolo:
I argue that the current behavior could be counter-intuitive. If users' customizations are discarded, I want that to be explicitly said. 
Also, if `vterm-keymap-exceptions` means "kaybindings that are bound to `nil`", the default value should change. In its current form, there are some inconsistencies: it has to be clear what this is about.

Finally, I want to think about is what is the role of this piece of code within the bigger picture.

wyuenho:
@Sbozzolo see how you like the new commits?

Sbozzolo:
What about:

```emacs-lisp
(defun vterm--key-bound-to-vterm-send-p (keybinding)
  "Return t if keybinding is bound to a function with a name that
starts with vterm-send-."
  ;; If keybinding is bound to a function with name vterm-send-* we want to return t,
  ;; otherwise nil.
  ;; (lookup-key vterm-mode-map (kbd keybinding)) has three possible outcomes
  ;; 1. nil, when keybinding is not bound in vterm-mode-map
  ;; 2. The symbol which is bound to keybinding
  ;; 3. A list (if keybinding is a prefix)
  ;;
  ;; In case 1 and 3, we will return nil.  In case 2., we will check the name of
  ;; the symbol.
  ;;
  ;; We need to add (kbd) so that the user can pass unescaped keys, e.g. "C-p"
  ;; instead of "\C-p"
  (let ((looked-up-key (lookup-key vterm-mode-map (kbd keybinding))))
    (when looked-up-key
      (unless (listp looked-up-key)
        (string-prefix-p "vterm-send-" (symbol-name looked-up-key))))))
```

And 
```emacs-lisp
(cl-loop for exception in vterm-keymap-exceptions
           if (vterm--key-bound-to-vterm-send-p exception)
           do (define-key vterm-mode-map (kbd exception) nil))
```

With this, you can change `vterm-keymap-exception` and this will take effect in new vterm session. The only restriction is that if users bound keys to commands with name `vterm-send-XXXX`, this will be removed.

What do you think?

wyuenho:
> The only restriction is that if users bound keys to commands with name vterm-send-XXXX, this will be removed.

I thought the whole intention of dragging this PR on is to not interfere with the user's decisions. This whole argument between whether the binding in the keymap or the exception defcustom wins when a vterm session is launched is about picking a winner for the user. This is never going to be "intuitive" unless what we picked by default is exactly the same as what the user thinks will happen. Perhaps all of this decisions shouldn't be made by us, but by the user after all. If you are thinking of yet another option to let the user choose which wins, I would advise against it as it can only make matters even more complicated than it already has.

I don't want to dwell on this any longer as the time spent on this bug fix is getting absolutely ridiculous. All I wanted to do is to fix an obvious bug in the exception unbinding process.

I maintain that either we shouldn't deal with this situation. The last two commits are a compromise, take it or leave it.

Issue #398, opened 2020-09-13
ipython: cursor not moved after inserting spaces at the end of the line


blahgeek:
How to reproduce:

- Open `ipython`
- Press space

Expected behavior: cursor moves to right

Actual behavior: cursor not moved

Sbozzolo:
I will close this issue as a duplicate of #196.

PR #399, opened 2020-09-13
Correctly place cursor after EOL

blahgeek:
If the cursor position is set beyond the EOL, insert spaces

Fixes https://github.com/akermu/emacs-libvterm/issues/398

---

Some application will want to place the cursor after EOL. E.g., in iPython, after pressing some spaces at the end of the line, the content of current line will be like "[0]:" (without any spaces) but the column of the position of the cursor may be 6 (greater than the length of the content of current line). This PR fixes this bug by inserting additional spaces at end of the line so that the cursor can be correctly placed.

Calls to `get_col_offset` and `forward_char` together is now replaced with one function `goto_col`, which will place the cursor at the target column (like `goto_line`). `goto_col` is pretty the same as `get_col_offset` except that it also calculates how many spaces there is after the EOL. After calling `forward_char` as before, it then inserts spaces if required.

### DIFF
--- a/vterm-module.c
+++ b/vterm-module.c
@@ -233,7 +233,7 @@ static int is_end_of_prompt(Term *term, int end_col, int row, int col) {
   return 0;
 }
 
-static void goto_col(Term *term, emacs_env *env, int row, int end_col) {
+static void goto_col(Term *term, emacs_env *env, int row, int end_col, bool fill_eol_spaces) {
   int col = 0;
   size_t offset = 0;
   size_t beyond_eol = 0;
@@ -259,9 +259,11 @@ static void goto_col(Term *term, emacs_env *env, int row, int end_col) {
   }
 
   forward_char(env, env->make_integer(env, end_col - offset));
-  emacs_value space = env->make_string(env, " ", 1);
-  for (int i = 0 ; i < beyond_eol ; i += 1)
-    insert(env, space);
+  if (fill_eol_spaces) {
+    emacs_value space = env->make_string(env, " ", 1);
+    for (int i = 0 ; i < beyond_eol ; i += 1)
+      insert(env, space);
+  }
 }
 
 static void refresh_lines(Term *term, emacs_env *env, int start_row,
@@ -473,7 +475,7 @@ static void adjust_topline(Term *term, emacs_env *env) {
    */
 
   goto_line(env, pos.row - term->height);
-  goto_col(term, env, pos.row, pos.col);
+  goto_col(term, env, pos.row, pos.col, true);
 
   emacs_value windows = get_buffer_window_list(env);
   emacs_value swindow = selected_window(env);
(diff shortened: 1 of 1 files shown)


Sbozzolo:
Can you please explain your changes?

blahgeek:
> Can you please explain your changes?

Sure, see the updated description.

schellj:
The fix in this PR doesn't quite work due to the fact that the vterm buffer is read-only, causing the `insert`s to throw `Buffer is read-only: #<buffer vterm>` errors. Possible solution here: https://github.com/schellj/emacs-libvterm/commit/5e7c76478e7215c6159e815427fa5aef2a52c847

blahgeek:
@schellj Thanks for your reply, but it worked well during my testing. Can you be more specific about how to reproduce the error?

schellj:
@blahgeek 

1. `M-x vterm`
2. Run `vi`
3. Enter insert-mode
4. Press space
5. Enter `vterm-copy-mode`
6. Exit `vterm-copy-mode`
7. Observe `Buffer is read-only: #<buffer vterm>` error

blahgeek:
OK I got it. I had never used the copy mode (I use evil). I pushed another possible solution to try fix this.

schellj:
Thanks, I'll test today.

schellj:
@blahgeek there are still issues with this fix.

1. M-x vterm
2. Run vi
3. Enter insert-mode
4. Press space
5. Enter vterm-copy-mode
6. Exit vterm-copy-mode
7. Observe that the cursor resets to the beginning of the line and cannot be moved to the right until more text is entered.

blahgeek:
@schellj Thanks again for your testing. However I cannot reproduce, the cursor remains at the end of the line after entering and exiting `vterm-copy-mode` in my case. But as I mentioned, I personally don't use copy mode so I may have missed something. I also tested your patch https://github.com/schellj/emacs-libvterm/commit/c56c90516b0fef776b044b3f54645c3d4c74aeec and can confirm that it's also working fine for me. I think I'm going to leave this pull request as is for the owner to decide which commit to merge or not.

schellj:
@blahgeek repro video if you're interested: https://vimeo.com/459441092
That was using Emacs version
```
GNU Emacs 27.1 (build 1, x86_64-apple-darwin19.6.0, NS appkit-1894.60 Version 10.15.6 (Build 19G73)) of 2020-09-04
```

Sbozzolo:
Thank you both. It seems that we are on the right track to solve this problem.

I will try to have a look at this in the weekend, but I have a long backlog, so I am not sure I'll be able to work on it.

jixiuf:
I cherry-pick 28ce5f6, and push another commit to make it work for` vterm-copy-mode`

schellj:
Thanks @jixiuf, your update appears to fix this issue, at least as far as I've seen so far.

Issue #362, opened 2020-07-10
M-/  (dabbrev-expand) requires buffer to be writable


sje30:
Hi,

if I use M-/  (dabbrev-expand) in a vterm buffer, I get the error:

```
command-execute: Buffer is read-only: #<buffer vterm<3>>
```

I'm relatively new to vterm (but not Emacs); I can see that toggling read-only with C-x C-q allows dabbrev-expand to work.  Is there any reason why read-only is on?  I'd regard dabbrev-expand has completion similar to TAB completion.

Sbozzolo:
The reason vterm buffers are read-only is that we have to properly send the keys to the underlying shell. Just writing on the buffer is not enough. 

I am surprised if `dabbrev-expand` works once you toogle read only. It may write something on screen, but I suspect that those character are not sent to the shell. Am I wrong?

sje30:
Thanks, and you were right.    Although it completes in the buffer, when I hit RET the completed text does not get sent to the shell.

Sbozzolo:
I am reopening this. It shouldn't be too difficult to implement a fix, but I haven't looked into this yet.

vindarel:
Hello, here's what I found. I miss the last steps though.

Disable read-only mode, use dabrrev, insert what was completed to vterm:

```elisp
    ;; user types "foo" and wants to complete:
    (vterm-insert (s-chop-prefix dabbrev--last-abbreviation dabbrev--last-expansion))
    ;;                                          ^^ foo                                  ^^ foo_bar  => append "_bar" to the existing "foo".   
```

This constructs a usable line for vterm.

I looked for dabbrev hooks to do this automatically after an expansion but couldn't find hooks.

I tried to re-bind `M-/` to a custom function, but I wasn't able to make it take precedence^^

```elisp
(defun my-vterm-dabbrev ()
         (call-interactively #'dabbrev-expand)
         (vterm-insert (s-chop-prefix dabbrev--last-abbreviation dabbrev--last-expansion)))

(define-key vterm-mode-map [M-/]    #'my-vterm-dabbrev)  ;; it still calls dabbrev-expand directly.
```

t0yv0:
https://github.com/t0yv0/t0yv0-emacs/blob/e8dfd5422105e77a7a66203c9cabfbd74e6e3c21/t0yv0-ware.el#L24 FWIW I've got something working for myself by dropping to lower-level primitives such as dabbrev--find-expansion.

port19x:
[...]

PR #401, opened 2020-09-16
Add vterm-dabbrev-expand

wigust:
This commit add dabbrev-expand as requested in #362


Sbozzolo:
Thanks for this PR. I am not sure if this is the best way to solve the problem. One issue I can see is that if one has autocompletion (e.g. in fish, or zsh with autocompletion) the thing at the point will be the entire command (even if part of it will be greyed out). 

I suspect that what will happen is: say one wants to autocomplete "python", having typed "py", `vterm-dabbrev-expand` will yield `pythonthon`. 

Possible ingredients we can use: `dabbrev--last-expansion` (where dabbrev is run in a cloned buffer), `vterm-undo`, `vterm-send-C-w`.

I don't have the solution, but I think we can find something more robust than your counting the number of characters and deleting them.

Can you please test `vterm-dabbrev-expand` on a terminal with autocompletion?

wigust:
Thanks for the review.  I applied your suggestions.  Now `vterm-dabbrev-expand` works greats in Fish shell with autocompletion.

Sbozzolo:
Thanks, I see some room for improvement, but overall I think the general strategy is good. Can you please explain why you need the check on `things-at-point` in the second function?

okamsn:
Could this be generalized to work with `completion-at-point`?

wigust:
@Sbozzolo `thing-at-point` is required in case you have a cursor after space, in which case `vterm-dabbrev-expand` completes a word after a matching word as `dabbrev-expand`, e.g. after space after "This" word the `vterm-dabbrev-expand` will complete `buffer`.
```
;; This buffer is for text that is not saved, and for Lisp evaluation.
...
```

@okamsn i don't see how `completion-at-point` could be used, because it only shows completion in a separate window.  Maybe we could have a `completion-at-point-functions`, but I don't see the point to do so to replace current implementation.

okamsn:
> @okamsn i don't see how `completion-at-point` could be used, because it only shows completion in a separate window.  Maybe we could have a `completion-at-point-functions`, but I don't see the point to do so to replace current implementation.

What I mean to ask is whether there is a way to make this insertion of text work for other commands, such as `completion-at-point`? I found code to complete file names at point, and it would be nice to use in Vterm.

wigust:
> What I mean to ask is whether there is a way to make this insertion of text generally work for other commands too, such as completion-at-point? I found code to complete file names at point, and it would be nice to use in Vterm.

completion-at-point and file name completion has nothing to do with dabbrev-expand and vterm-dabbrev-expand which tries to mimic dabbrev-expand.

Issue #403, opened 2020-09-19
libvterm>0.1.3 VTermStringFragment breaks emacs-libvterm


Sbozzolo:
In 1d7d72b63b21eb0d5a66da3163bc2c2b59bc3908, the developers of libvterm introduced a new way to handle strings with a struct `VTermStringFragment`. We are still using the previous infrastructure with plain arrays of chars.

While most users are using libvterm<=0.1.3, it is important to keep up with libvterm, so that we can enjoy the improvements and bugfixes in the library. This change is not too big, but it is not simple to handle because we need to maintain backward compatibility. I don't know how to do this. (My only idea is to have a branch that supports libvterm>0.1.3, but this is not ideal.)

wyuenho:
Maybe dropping support for system libvterm is a good idea. The vterm-module requires compilation and build essentials to exist anyway, why bother supporting old versions?

condy0919:
vterm failed to start, is it related?

```
Debugger entered--Lisp error: (error "Memory exhausted--use C-x s then exit and restart ...")
  vterm--new(45 172 1000 nil nil nil)
  vterm-mode()
  vterm()
  funcall-interactively(vterm)
  call-interactively(vterm nil nil)
  command-execute(vterm)
```

condy0919:
> vterm failed to start, is it related?
> 
> ```
> Debugger entered--Lisp error: (error "Memory exhausted--use C-x s then exit and restart ...")
>   vterm--new(45 172 1000 nil nil nil)
>   vterm-mode()
>   vterm()
>   funcall-interactively(vterm)
>   call-interactively(vterm nil nil)
>   command-execute(vterm)
> ```

Looks like an emacs bug, has been fixed now.

PR #404, opened 2020-09-21
upgrade libvterm

### LINKED ISSUE #389: compilation error on mac
I was getting compilation error when trying to compile libvterm on my machine:

macOS Catalina version 10.15.6
had installed cmake and libtool with brew before compilation

image

I don't know if there's a better way to solve this, but my workaround solution was to add 
`#define _POSIX_C_SOURCE 200809L`
on top of `vterm-ctrl.c` file
I got this solution from https://stackoverflow.com/questions/26284110/strdup-confused-about-warnings-implicit-declaration-makes-pointer-with

I was wondering if any other people had similar issues trying to compile on mac and if there was a better solution. Also, was wondering if the source code can be updated.

jixiuf:
#403 
should fix  #389 

Edit: should ready to be reviewed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,11 @@ if (USE_SYSTEM_LIBVTERM)
   # vterm.h is found.
   if (LIBVTERM_INCLUDE_DIR)
     message(STATUS "System libvterm detected")
+    execute_process(COMMAND  grep -c "VTermStringFragment" "${LIBVTERM_INCLUDE_DIR}/vterm.h" OUTPUT_VARIABLE VTermStringFragmentExists)
+    if (${VTermStringFragmentExists} EQUAL "0")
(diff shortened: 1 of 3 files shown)


cwhatley:
FYI, I tried this branch on emacs 27.1/mac catalina and haven't run into any issues.

Sbozzolo:
Thank you very much! That's great. I'll have a look.

Sbozzolo:
I haven't looked at this in depth yet, but I think we should finish #410 first. This new PR makes the build system more fragile, so it would be great to have some form of continuous integration to run some smoke tests.

olreim:
Works great. No issues. Emacs 27.1 running Doom 2.0.9 on macOS Catalina 10.15.6.

jsigman:
I can currently only get vterm to work with this branch, using a Mac OS with the latest software and updates. It also runs on both 27.1 and also on the native-comp branch

PR #405, opened 2020-09-22
avoid garbling of certain multibyte characters .

### LINKED ISSUE #394: Mojibake when output non-ASCII Unicode Characters on certain conditions
If the 1024th(1-based index) byte of a process's output is any but the last byte of  of a Unicode Character(non-ASCII) bytes sequence, e.g.:

1. OK
```shell
ruby -e 'print "x"*1020+"😀"'
```

2. Mojibake
```shell
ruby -e 'print "x"*1021+"😀"'
```

3. Mojibake
```shell
ruby -e 'print "x"*1022+"😀"'
```

4. Mojibake
```shell
ruby -e 'print "x"*1023+"😀"'
```

5. OK
```shell
ruby -e 'print "x"*1024+"😀"'
```

ps: UTF-8 byte sequence of 😀 is of four bytes

jixiuf:
should fix #394 .

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -430,6 +430,7 @@ Only background is used."
 (defvar-local vterm--redraw-immididately nil)
 (defvar-local vterm--linenum-remapping nil)
 (defvar-local vterm--prompt-tracking-enabled-p nil)
+(defvar-local vterm--undecoded-bytes nil)
 
 (defvar vterm-timer-delay 0.1
   "Delay for refreshing the buffer after receiving updates from libvterm.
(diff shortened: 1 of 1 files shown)


Sbozzolo:
Thanks, this is neat. Looks good to me, if you can please reformat the comment as follows
```
        ;; Borrowed from term.el
        ;;
        ;; Avoid garbling of certain multibyte characters by decoding the string
        ;; before counting characters.  See,
        ;; https://github.com/akermu/emacs-libvterm/issues/394, and the
        ;; https://debbugs.gnu.org/cgi/bugreport.cgi?bug=1006 (for term.el).
```
feel free to merge.

jixiuf:
done

PR #424, opened 2020-10-21
Add Cmake rules for OpenBSD & NetBSD

fagg:
BSD make barfs on the libvterm Makefile that gets pulled in if you try to build without the system libvterm (i.e. `cmake -DUSE_SYSTEM_LIBVTERM=no ..`) This adds rules for OpenBSD and NetBSD so that they use GNU make.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ include(ExternalProject)
 
 project(emacs-libvterm C)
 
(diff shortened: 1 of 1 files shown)


Sbozzolo:
Thanks, did you test this on those platforms?

fagg:
Tested on OpenBSD. I don't have NetBSD available.

Issue #321, opened 2020-05-09
Buffer face doesn't inherit from default face 


xofyarg:
I use "InputMonoNarrow Light" as default font by setting the ":font" property of the default face, but seems vterm uses the "InputMonoNarrow" variant. Because the weight is different, it makes me feel all the texts are bold.

Not sure if this is a bug or not, please let me know if I missed anything.

Sbozzolo:
Vterm should inherit from the default face. Can move the point to a character in vterm and press `C-u C-x =`? It will tell you what font is being used.

xofyarg:
Yeah, that's how I know the font weights are different. My guess is vterm just set it to normal no matter what weight default face is using.

Sbozzolo:
I think I found the bug. We have a switch to disable bold fonts, this sets the weight of a face to be normal or bold and it disregards the information about the weight in the default face.

This is in the function `render_text` in the `vterm-module.c`

Not sure how to fix, but it is a bug.

PR #429, opened 2020-10-25
Minimize text properties

heytings:
emacs-libvterm creates lots of text properties on each character. I propose to minimize them. In particular, with this patch `vterm--get-color` can return nil when the default face should be used, which is in general the case for most characters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,7 +459,7 @@ C-c`.
 
 ## Fonts
 
-If you would like to change the font or face used in a vterm, use the following code:
+You can change the font (the _face_) used in a vterm with the following code:
 
 ``` emacs
 (add-hook 'vterm-mode-hook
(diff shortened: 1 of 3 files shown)


Sbozzolo:
Thanks!

I did some cleanup. In particular, if we are going to check for `bold` to be `Qnormal` to ignore it, might as well make it `Qnil`, so that all the variables are handled in the same way.

This should fix #321.

heytings:
Thanks!
BTW, it's not clear to me why ":extend t" is set on all characters with Emacs >= 27.  I tried to disable it, and did not experience any problems.  What's the intended purpose of this property?
Another note: perhaps `if (props_len) put_text_property(env, text, Qface, properties);` in `render_text()` would make sense.

Sbozzolo:
In Emacs 27, if you want a text property to continue until the end of the line, you have to set `:extend` to `t`. For example, with `htop`, if you want to have it nicely rendered, you need it. If you disable it, I would expect `htop` to mis-render.

heytings:
Yes, I know this.  But it's not clear to me that it's necessary to put that property it on each inserted text, it could be put only on the last text of each line.
I tried `top` which works okay without `:extend t`, but indeed `htop` does not work well without it.

Sbozzolo:
Yes, I agree with you, and I don't know what it is for every character. I suppose the reason is that it is much simpler to do that than to put it only at the end on the line.

heytings:
About b2e631c: IMO one could also return `nil` for -11 and -12, and remove the `vterm-color-underline` and `vterm-color-inverse-video` faces.  Or perhaps add an option to use them, and return nil when they are not used.

PR #430, opened 2020-10-25
Disable all automatic fontification

### LINKED ISSUE #425: Use font-lock-defaults?
I discovered this by chance: https://github.com/akermu/emacs-libvterm/issues/46#issuecomment-451921161

Which was the fix I needed for my vterm-buffers-turning-green problem.

However this didn't work as expected:

```elisp
(use-package vterm
  ...
  :hook
  (vterm-mode . (lambda ()
                  (setq-local font-lock-keywords-only t))))
```
(`font-lock-keywords-only` would be nil when I checked)

After some digging about, I found out this worked:

```elisp
(use-package vterm
  ...
  :hook
  (vterm-mode . (lambda ()
                  (setq-local font-lock-defaults '('() t))))
```

`font-lock-keywords-only` is actually set from the value of `font-lock-defaults`. Which seems to happen on the first fontification, which clearly was run after the mode-hook in my case.

How about adding `(setq-local font-lock-defaults '('() t))` to the mode?

Sbozzolo:
Fixes #425

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -572,6 +572,10 @@ Exceptions are defined by `vterm-keymap-exceptions'."
     (setq-local hscroll-margin 0)
     (setq-local hscroll-step 1)
     (setq-local truncate-lines t)
+
+    ;; Disable all automatic fontification
+    (setq-local font-lock-defaults '(nil t))
+
     (add-function :filter-return
                   (local 'filter-buffer-substring-function)
                   #'vterm--filter-buffer-substring)


KaratasFurkan:
Thanks to this PR, I could manage to fontify docker container prompts in vterm.

Screenshot_20201109_181652

Code: https://github.com/KaratasFurkan/.emacs.d/commit/32bf827da2c7b6c4663c54423e6ce46e1596410b#diff-ecbc1aa90e9ff97a00b0b2aab1551bceee0c4d21993146bdcb1af4de31c9cac6R3018-R3028

PR #460, opened 2020-12-05
vterm-yank: fix wrong symbol-function call

caribpa:
`vterm-yank` broke after the commit https://github.com/akermu/emacs-libvterm/commit/f21d0dd1abc26ae447736c978cb2ea942627cd90 due to wrong `symbol-function` call (`insert` instead of `insert-for-yank`)

This pull request fixes the issue 🏆

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -920,7 +920,7 @@ Argument ARG is passed to `yank'."
   (interactive "P")
   (vterm-goto-char (point))
   (let ((inhibit-read-only t))
-    (cl-letf (((symbol-function 'insert) #'vterm-insert))
+    (cl-letf (((symbol-function 'insert-for-yank) #'vterm-insert))
       (yank arg))))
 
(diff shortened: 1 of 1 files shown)


jixiuf:
weird, `insert` work for me on emacs 28, I will merge your pull request first.

caribpa:
Well, I am on emacs 27.1 so that may explain it.

To be double sure I tried without my changes in `emacs -Q` (forgot to do that before) and the issue is still there.

Just to be triple sure, are you certain you didn't map `C-y` to `vterm--self-insert` instead of the default `vterm-yank`? Because the shell has its own clipboard and its content can be pasted using `C-y` (`C-c C-y` in vterm.el), which works as expected without my changes.

jixiuf:
I am sure `vterm-yank` is called .

caribpa:
Then something is different in emacs 28 vs emacs 27.1

PR #473, opened 2021-01-04
fix rendering of combining characters

Dieken:
Some emoji characters are composed by some other emoji characters with
zero width joiner, such as:
🏊🏻‍♀️
🏊🏻‍♂️
👩‍👩‍👧
👨‍👨‍👧‍👦

With this patch, these emoji characters can be correctly rendered by
https://bitbucket.org/mituharu/emacs-mac.


Dieken:
@akermu  @jixiuf  could you take a look?

PR #475, opened 2021-01-08
Fix vterm_printf tmux escape sequence pass through

### LINKED ISSUE #443: Issue with the shell side config.
Hi:

I see that the vterm_printf function has an special case when TMUX is defined... I am not sure if it refers to the case of calling tmux from vterm or when calling vterm from an emacs inside a tmux.

and the link in the file: http://permalink.gmane.org/gmane.comp.terminal-emulators.tmux.user/1324

is broken.

When I do:

```
tmux
emacs
M-x vterm
```

I get this prompt:

image

I made a work around with something like this and it solved the issue for the moment:

```
if [[ "$INSIDE_EMACS" = 'vterm' ]] && ...; then
 	source ${EMACS_VTERM_PATH}/etc/emacs-vterm-bash.sh
	export TMUX=
fi
```

I know that calling vterm from emacs from tmux looks like a difficult scenario; but sometimes it is useful when working with different files and I need a quick terminal in the current path (vterm-toggle comes to rescue)... 

On the other hand tmux does not import variables like INSIDE_EMACS unless they are added in the tmux option: `update-environment` so this makes me wonder when is useful that if TMUX condition?

Best,
Ergus

felker:
tmux requires that `TERM` is set to `screen*` or `tmux*` for subshells running inside tmux. See https://github.com/tmux/tmux/wiki/FAQ

Using this, I was able to change the `vterm_printf` passthrough switch such that it is not activated if Emacs and vterm are running **within** a tmux session. The passthrough still functions for running tmux within Emacs and vterm.

I have not yet tested the changes to the Fish shell-side configuration. 

Closes #443.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ readme.
 For `bash` or `zsh`, put this in your `.zshrc` or `.bashrc`
 ```bash
 vterm_printf(){
-    if [ -n "$TMUX" ]; then
+    if [ -n "$TMUX" ] && ([ "${TERM%%-*}" = "tmux" ] || [ "${TERM%%-*}" = "screen" ] ); then
         # Tell tmux to pass the escape sequences through
         printf "\ePtmux;\e\e]%s\007\e\\" "$1"
     elif [ "${TERM%%-*}" = "screen" ]; then
@@ -258,7 +258,7 @@ This works also for `dash`.
 For `fish` put this in your `~/.config/fish/config.fish`:
 ```bash
 function vterm_printf;
-    if [ -n "$TMUX" ]
+    if begin; [  -n "$TMUX" ]  ; and  string match -q -r "screen|tmux" "$TERM"; end 
         # tell tmux to pass the escape sequences through
         printf "\ePtmux;\e\e]%s\007\e\\" "$argv"
     else if string match -q -- "screen*" "$TERM"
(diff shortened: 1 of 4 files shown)


Sbozzolo:
I suspect I am missing one piece. In `vterm` we set `TERM` to `xterm-256colors` by defaults. Does `tmux` somehow override that, or do you have to do it manually?

felker:
In this branch, `vterm` continues to set `TERM` to `xterm-256colors` by default, even for Emacs inside of tmux. This is good, since it now prevents the first conditional branch of `vterm_printf` from executing

```
if [ -n "$TMUX" ] && ([ "${TERM%%-*}" = "tmux" ] || [ "${TERM%%-*}" = "screen" ] ); then
        # Tell tmux to pass the escape sequences through
        printf "\ePtmux;\e\e]%s\007\e\\" "$1"
```
(since we only want that passthrough sequence for tmux inside vterm, not vterm inside tmux).

`tmux` sets `TERM` to `screen`, by default. But I (and many others) have set it to `tmux-256colors` in my `~/.tmux.conf` via:
```
set -g default-terminal "tmux-256color"
```

From tmux docs:
> PLEASE NOTE: most display problems are due to incorrect TERM! Before
reporting problems make SURE that TERM settings are correct inside and
outside tmux.
> Inside tmux TERM must be "screen", "tmux" or similar (such as "tmux-256color").
Don't bother reporting problems where it isn't!
> Outside, it should match your terminal: particularly, use "rxvt" for rxvt
and derivatives.

PR #478, opened 2021-01-09
fix cursor type after hide cursor and show again

Dieken:
This patch also adds support for cursor blink.

VIM emits these ANSI escape sequences in insert mode:
```
    \x1b[?25h   show cursor
    \x1b[?25l   hide cursor
```

You may add these to ~/.vimrc to verify it:
```
" https://vim.fandom.com/wiki/Change_cursor_shape_in_different_modes
" Cursor settings:
"  1 -> blinking block
"  2 -> solid block 
"  3 -> blinking underscore
"  4 -> solid underscore
"  5 -> blinking vertical bar
"  6 -> solid vertical bar
let &t_SI="\e[5 q"
let &t_SR="\e[3 q"
let &t_EI="\e[1 q"
```

@jixiuf

### DIFF
--- a/README.md
+++ b/README.md
@@ -475,6 +475,13 @@ In addition to that, you can disable some text properties (bold, underline,
 reverse video) setting the relative option to `t` (`vterm-disable-bold`,
 `vterm-disable-underline`, or `vterm-disable-inverse-video`).
 
+## Blink cursor
+
+When `vterm-ignore-blink-cursor` is `t`, vterm will ignore request from application to turn on or off cursor blink.
+
+If `nil`, cursor in any window may begin to blink or not blink because `blink-cursor-mode`
+is a global minor mode in Emacs, you can use `M-x blink-cursor-mode` to toggle.
+
 ## Colors
 
 Set the `:foreground` and `:background` attributes of the following faces to a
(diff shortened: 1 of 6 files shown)


jixiuf:
`blink-cursor-mode` is a global minor-mode,

Dieken:
> `blink-cursor-mode` is a global minor-mode,
> #98
> #104

No idea why Emacs doesn't have buffer local blink cursor mode :-(

If the application running in vterm doesn't use blink cursor,   this patch won't change blink cursor mode, so I feel it's acceptable:  it does that application demands, user can call M-x blink-cursor-mode to toggle it if needed.

jixiuf:
somebody prefer disable  `blink-cursor-mode`.

Emacs -Q 
```
(blink-cursor-mode -1)
```
 blink-cursor-mode should be disabled now 

then  open vi in vterm buffer ,and quit vi

`blink-cursor-mode` is enabled now.  Somebody don't like this(vterm change the behavior of `blink-cursor-mode`)

Dieken:
VIM doesn't set t_SI etc by default,  anyway,  let's add an defvar `vterm-enable-support-blink-cursor`.

Dieken:
@jixiuf   I added a customization variable `vterm-ignore-blink-cursor`, default to `t`,  so that vterm won't touch blink status by default.

Dieken:
Any more comment? @jixiuf

PR #484, opened 2021-01-18
Put module compilation buffer in compilation mode

wyuenho:
So the buffer is quittable by pressing q, and the compilation warnings and errors are navigable


Sbozzolo:
Thanks!

PR #485, opened 2021-01-20
Fix xterm-paste in vterm.

JimDBh:
Vterm-yank does not copy from the system clipboard that is supported
via xterm.el. This patch fixes it.


tarsius:
I've added a fix for this in #485.

PR #536, opened 2021-07-11
Correctly append to `PROMPT_COMMAND` for bash.

johnmwu:
If `PROMPT_COMMAND` is empty, the original would leave `;vterm_set_directory` as its value, giving error

```
[michael@contessa emacs-libvterm]$ PROMPT_COMMAND=";vterm_set_directory"
bash: PROMPT_COMMAND:16: syntax error near unexpected token `;'
bash: PROMPT_COMMAND:16: `;vterm_set_directory'
```


Sbozzolo:
Thanks, can you also please update the files in the `etc` folder?

PR #537, opened 2021-07-12
Support custom `vterm-copy-mode-map`

schellj:
Not sure if this is correct or required, but without this change, I haven't been able to figure out a way to apply a customized `vterm-copy-mode-map`.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -772,7 +772,7 @@ A conventient way to exit `vterm-copy-mode' is with
   (if (equal major-mode 'vterm-mode)
       (if vterm-copy-mode
           (progn                            ;enable vterm-copy-mode
-            (use-local-map nil)
+            (use-local-map vterm-copy-mode-map)
             (vterm-send-stop))
         (vterm-reset-cursor-point)
         (use-local-map vterm-mode-map)


Sbozzolo:
Admittedly, why we are setting the local map to nil, but if I evaluate the following code:
```emacs-lisp
(defun say-hi ()
  (interactive)
  (message "HI"))

(define-key vterm-copy-mode-map (kbd "C-p C-f") 'say-hi)
```
`C-p C-f` correctly prints "HI" to the minibuffer when I am in `vterm-copy-mode`. Is this what you are trying to achieve?

schellj:
@Sbozzolo yes, that method seems to work, thanks.

I also just tested and doing the following in `use-package` `:init` works, but not in `:config`: 
```
  (setq-default vterm-copy-mode-map (let ((map (make-sparse-keymap)))
                              (define-key map (kbd "M-p")      #'vterm-copy-mode)
                              (define-key map [return]         #'vterm-copy-mode)
                              (define-key map (kbd "RET")      #'vterm-copy-mode)
                              (define-key map (kbd "H-c C-r")  #'vterm-reset-cursor-point)
                              (define-key map (kbd "C-a")      #'vterm-beginning-of-line)
                              (define-key map (kbd "C-e")      #'vterm-end-of-line)
                              (define-key map (kbd "H-c C-n")  #'vterm-next-prompt)
                              (define-key map (kbd "H-c C-p")  #'vterm-previous-prompt)
                              (define-key map (kbd "H-c C-f")  #'ffap)
                              map))
```

I'm fairly certain that I had tried that before and had issues, but I can't seem to reproduce them now. Possibly something changed in `emacs-libvterm` since July that enabled setting `vterm-copy-mode-map` in `use-package` `:init` to work, not sure.

PR #552, opened 2021-11-02
Add nested tmux escape sequence for bash prompt tracking

d-p:
The single tmux instances are unaffected by the change.

The implementation uses the single quote to simplify the escaping of
backslashes (avoid doubling them) and replaces the octal value of the
bell sequence (\007) with the corresponding C escape sequence (\a) for
uniformity with ESC (\033 or \e).

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,8 +243,12 @@ For `bash` or `zsh`, put this in your `.zshrc` or `.bashrc`
 ```bash
 vterm_printf(){
     if [ -n "$TMUX" ] && ([ "${TERM%%-*}" = "tmux" ] || [ "${TERM%%-*}" = "screen" ] ); then
-        # Tell tmux to pass the escape sequences through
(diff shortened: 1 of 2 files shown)


Sbozzolo:
Thanks for your PR. Is there a reason to prefer the "simple" method as opposed to the one that works for nested sessions as well?

Sbozzolo:
I don't know much about tmux, but we should definitely ensure that the same behavior is available for zsh and fish as well (ie, we should also edit the corresponding files). If everything is consistent, we can merge this, leaving a descriptive comment. The comment should explain why one might want to use the simple version (which is more or less what you wrote in the last message). 

Thanks.

d-p:
Thanks for the quick feedback loop, I will try to add some tests and then add/refactor the shell code.
Probably the tests will be another pull request (hopefully in the near future) and then I will update this one.

Issue #559, opened 2021-12-01
Seg fault on emacs-28 with native-comp


Dadams2:
This is probably going to be an absolute extreme edge case. I am on a HPC system still using centos7 so to get and up-to-date emacs I have had to compile and link everything myself. I suspect this is going to be very hard to reproduce so if there is an obvious thing that I have done wrong let me know.

gcc 12 built from source bootstraped with gcc 9.2.0 with the following options:
```
../src/configure --enable-host-shared --enable-languages=c,c++,jit  --enable-checking=release --prefix=$HOME/local --with-gmp=$HOME/local --with-mpfr=$HOME/local --with-mpc=$HOME/local --disable-multilib
```
emacs-28 branch (30553d889d) built with said gcc 12 with the following options:
```
./configure --with-modules --prefix=/node/bud82/davida/attempted-builds/emacs27 --x-includes=/node/bud82/davida/envs/emacs/include --x-libraries=/node/bud82/davida/envs/emacs/lib --with-x-toolkit=gtk3 --with-harfbuzz -with-cairo 'CFLAGS=-march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /node/bud82/davida/envs/emacs/include' 'CPPFLAGS=-DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /node/bud82/davida/envs/emacs/include' 'LDFLAGS=-Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/node/bud82/davida/envs/emacs/lib -Wl,-rpath-link,/node/bud82/davida/envs/emacs/lib -L/node/bud82/davida/envs/emacs/lib' --with-jason --with-native-compilation
```
envs/emacs is a place where I compiled and installed all the required dependencies for emacs native comp. **I do not get the seg fault when I build the emacs-27.2 release in the exact same way. Vterm functions correctly in this instance**.  I took a lot of inspiration from the conda installation of emacs for my compilation options as they look sane for a custom build environment. 

I am running the latest doom emacs but compiled vterm-module.so myself with the following:
```
mkdir build && cd build
cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
make -j20
```
I get the following seg fault any time I try to open vterm: 
```
[...]

PR #562, opened 2021-12-14
Do not export unnecessary symbols in dynamic library

blahgeek:
Fixes https://github.com/akermu/emacs-libvterm/issues/559

As commented in https://github.com/akermu/emacs-libvterm/issues/559#issuecomment-992681062, emacs crashes when loading vterm-module.so because the address of `Qt` is incorrectly resolved to the read-only symbol in `emacs` binary. I can verify this with `LD_DEBUG=all`:

```
     75570:     binding file .../.emacs.d/straight/build/vterm/vterm-module.so [0] to ./src/emacs [0]: normal symbol `Qt'
```

This PR hides all symbols except two necessary ones, so that the symbols like `Qt` will be resolved correctly.

I'm still unclear about the root cause: I thought `RTLD_LOCAL` should be the default behavior for `dlopen` so this should not happen; It's also strange that this only happens when building with `--with-xwidgets`, maybe loading xwidgets somehow affects the order of the symbol resolution. But anyway, this fix works and I don't think if there's any downside of it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ endif()
 add_library(vterm-module MODULE vterm-module.c utf8.c elisp.c)
 set_target_properties(vterm-module PROPERTIES
   C_STANDARD 99
+  C_VISIBILITY_PRESET "hidden"
   POSITION_INDEPENDENT_CODE ON
   PREFIX ""
   LIBRARY_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR}
--- a/vterm-module.h
+++ b/vterm-module.h
@@ -6,7 +6,22 @@
 #include <stdbool.h>
 #include <vterm.h>
 
-int plugin_is_GPL_compatible;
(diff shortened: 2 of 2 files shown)


Sbozzolo:
Thanks! I can merge this PR. However, I am rather ignorant about symbol visibility, so you could educate me on the topic providing more background for me to understand this commit, that would be very much appreciated. 

Thanks again.

blahgeek:
You may find more information in gcc document https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes  (search for "visibility_type").

Basically, "default" visibility exports the symbols so that they can be used (and may be overriden) outside of the shared library; while "hidden" visibility does not export the symbols, which means that they can only be used inside the shared library.

In this PR, I modified the preset visibility from "default" to "hidden" in cmake file (which essentially adds `-fvisibility=hidden` to compiler flags), and add function attribute to `emacs_module_init` and `plugin_is_GPL_compatible`, so that only these two symbols are exported.

PR #576, opened 2022-02-13
add LDFLAGS to statically link libvterm

### LINKED ISSUE #333: Cannot compile: `No rule to make target 'libvterm-prefix/src/libvterm/.libs/libvterm.a'`
Following the guide of manual installation,
```
❯ make
[  8%] Performing update step for 'libvterm'
[ 16%] No configure step for 'libvterm'
[ 25%] Performing build step for 'libvterm'
[ 33%] No install step for 'libvterm'
[ 41%] Completed 'libvterm'
[ 66%] Built target libvterm
make[2]: *** No rule to make target 'libvterm-prefix/src/libvterm/.libs/libvterm.a', needed by '../vterm-module.so'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:153: CMakeFiles/vterm-module.dir/all] Error 2
make: *** [Makefile:104: all] Error 2
```

xofyarg:
libvterm defaults to generate a dynamic library, however the emacs
module prefers a static one.

Fix #333


Sbozzolo:
Can you provide some references/more information, on why we should prefer static linking?

xofyarg:
Yeah, it basically fixes the issue mentioned in #333, that when working with the out-of-tree libvterm source, this module [expects a statically linked library](https://github.com/akermu/emacs-libvterm/blob/a940dd2ee8a82684860e320c0f6d5e15d31d916f/CMakeLists.txt#L76). Guessing it used to work fine but upstream has changed to produce a dynamically linked version now.

Thaodan:
IMHO then module should link libvterm dynamically if possible.
There's also ready the option to do.

ngasull:
I confirm this fixed #333 for me too :+1:

PR #581, opened 2022-03-16
fix: add an INSIDE_EMACS guard to vterm_prompt_end

evgeny-roubinchtein:
Without the guard, the shell prompt in buffers that are _not_ `vterm` ends up being messed up.
For example, if I do `M-x shell`, the shell prompt is messed up.


evgeny-roubinchtein:
Never mind: if people follow your advice, the `source`-ing of that entire file is will happen only if `[[ "$INSIDE_EMACS" = "vterm" ]]`, so this is actually the one place that _doesn't_ need the guard, because the guard is redundant here.

What _does_ need the guard is the snipped in the README -- but that's a different pull request.

PR #582, opened 2022-03-16
fix: add INSIDE_EMACS guards for vterm_prompt_end

evgeny-roubinchtein:
Add `INSIDE_EMACS` guard to `vterm_prompt_end` to avoid messing up the prompt for shells that are _not_ running in `vterm` (e.g., started via `M-x shell`).

### DIFF
--- a/etc/emacs-vterm-zsh.sh
+++ b/etc/emacs-vterm-zsh.sh
@@ -17,9 +17,7 @@ function vterm_printf(){
 
 # Completely clear the buffer. With this, everything that is not on screen
 # is erased.
-if [[ "$INSIDE_EMACS" = 'vterm' ]]; then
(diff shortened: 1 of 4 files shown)


etc/emacs-vterm-zsh.sh
```diff
@@ -17,9 +17,7 @@ function vterm_printf(){
 
 # Completely clear the buffer. With this, everything that is not on screen
 # is erased.
-if [[ "$INSIDE_EMACS" = 'vterm' ]]; then
-    alias clear='vterm_printf "51;Evterm-clear-scrollback";tput clear'
-fi
+alias clear='vterm_printf "51;Evterm-clear-scrollback";tput clear'
```
matthewbauer, line 20: I don't think you want this because it will overwrite /usr/bin/clear if you have that. Although `tput clear` might be enough on most systems?

PR #585, opened 2022-04-19
Add send-next-key function

tylergrinn:
Read next input event and send it to the libvterm.

This is useful for controlling an emacs session within emacs-libvterm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -455,6 +455,18 @@ or remove it from `vterm-mode-map`. By default, `vterm.el` binds most of the
 like `<backspace>` and `<return>`. Sending a keyboard interrupt is bound to `C-c
 C-c`.
 
+In order to send a keypress that is already recognized by Emacs, such as `C-g`,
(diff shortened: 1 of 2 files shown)


Sbozzolo:
How do you use this?

tylergrinn:
It can be called interactively or you can bind it:

`(define-key vterm-mode-map (kbd "C-q") #'vterm-send-next-key)`

So that `C-q C-g` sends a `C-g`

Sbozzolo:
Okay, it makes sense and I like it. Can you please, add documentation to explain the point of this function (in the readme as well).

tylergrinn:
Sure thing

Sbozzolo:
Great, thanks! Two nitpicks, then I'll merge this.
- `emacs` -> `Emacs`
- Can you add an example in the docstring? E.g., something like "with this you can directly send modified keys to applications running in the terminal (like Emacs or Nano)".

tylergrinn:
No problem!

PR #591, opened 2022-05-04
Declare variable before first use

tarsius:
Else the byte-compiler gets worried.


jixiuf:
I think this `PR` fix 
image

PR #612, opened 2022-08-06
Enable selection for commands in copy-mode

Inc0n:
This commit enables C-a and C-e to be used with shift modifer to select region to start of line and end of line, if `shift-select-mode` is enabled.

This is done by putting "^" into interactive, for more information see, https://www.emacswiki.org/emacs/ShiftSelectMode.

Additionally, I would recommend to bind M-w (king-ring-save) to `vterm-copy-mode-done` also, which is the default key bind to region save, this change I haven't introduce in this commit. I use this bind all the time, and had to manually quite copy-mode using C-c C-t, instead of using return.

I mentioned the above here, since this commit is about introducing better integration between vterm and emacs.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -1692,7 +1692,7 @@ More information see `vterm--prompt-tracking-enabled-p' and
 Move the point to the first character after the shell prompt on this line.
 If the point is already there, move to the beginning of the line.
 Effectively toggle between the two positions."
-  (interactive)
+  (interactive "^")
   (if (vterm--at-prompt-p)
       (goto-char (vterm--get-beginning-of-line))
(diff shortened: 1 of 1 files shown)


Inc0n:
This pull request is transfered to https://github.com/akermu/emacs-libvterm/pull/614 instead.

Sbozzolo:
I was not aware of `shift-select-mode`. I don't see any reason why we shouldn't merge this PR.

Inc0n:
Oh yes, `shift-select-mode`, I missed this one out (was actually not aware of this either), so this commit would be adding support for `shift-select-mode`. I have edited that into my first comment.

PR #619, opened 2022-08-20
Prefer vterm--self-insert

jixiuf:
prefer `vterm--self-insert`, and delete recent added `vterm-send-meta-{up/down/left/right}`


clhodapp:
This PR has been the bane of my existence since it was merged.

It destroyed a BUNCH of functions to allow calling key combos but didn't add them to the default binding set to restore their functionality. Specifically, vterm-send-C-a through C-z, C-A through C-Z, M-a through M-z, M-A through M-Z, and C-S-a through C-S-z have all been lost. As a result, I now have to manually bind new key chords each time I end up discovering that I need one of them.

It should really have bound all of the above combos underneath C-c if it was going to take them away.

PR #620, opened 2022-08-20
delete all unused command like vterm-send-C-[a-z] vterm-send-M-[a-z].

jixiuf:
After #619  is merged, commands like  `vterm-send-C-[a-z]` `vterm-send-M-[a-z]` and 
`vterm-send-{up/down/left/right/prior/next/meta-dot/meta-comma/ctrl-slash}` are useless for vterm.el.

But it is a breaking change for other packages,  so I am not sure should I merge this PR or not.

@Sbozzolo  @Inc0n

### DIFF
--- a/README.md
+++ b/README.md
@@ -848,6 +848,12 @@ not appropriate in some cases like terminals."
 
 Obsolete variables will be removed in version 0.1.
 
+#### August 2022
+* `vterm-send-C-[a-z]` `vterm-send-M-[a-z]` `vterm-define-key`
+ `vterm-send-{up/down/left/right/prior/next/meta-dot/meta-comma/ctrl-slash}`
+ were obsolete, please use `vterm-send` or `vterm-send-key` or
+ `vterm--self-insert` instead.
+
 #### October 2020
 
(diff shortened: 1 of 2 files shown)


Inc0n:
Since it is clear that vterm is of alpha stage, and breaking changes such as this is introduced from time to time.
But to take care into not breaking depending packages, `define-obsolete-function-alias` can be used, and phase them (remove them) out in a future release, as you see fit.

clhodapp:
These may not be useful for vterm internally, but they are extremely useful as commands: They allows you to e.g. easily send `c-x` to the process inside of vterm.

clhodapp:
This PR has been the bane of my existence since it was merged.

It destroyed a BUNCH of functions to allow calling key combos but didn't add them to the default binding set to restore their functionality. Specifically, vterm-send-C-a through C-z, C-A through C-Z, M-a through M-z, M-A through M-Z, and C-S-a through C-S-z have all been lost. As a result, I now have to manually bind new key chords each time I end up discovering that I need one of them.

It should really have bound all of the above combos underneath C-c if it was going to take them away.

PR #634, opened 2022-10-17
Use `prompt_pwd` to format `/home/myuser/mydir` as `~/mydir`

qst0528:
This commit improves documentation for [vterm-buffer-name-string](https://github.com/akermu/emacs-libvterm#vterm-buffer-name-string).
Fish has a utility function [`prompt_pwd`](https://fishshell.com/docs/current/cmds/prompt_pwd.html), which prints pwd suitable for prompt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,7 +434,7 @@ For `fish`,
 function fish_title
     hostname
     echo ":"
-    pwd
+    prompt_pwd
 end
 ```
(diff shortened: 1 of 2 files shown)


jixiuf:
Maybe this line should be updated too.
https://github.com/akermu/emacs-libvterm/blob/master/etc/emacs-vterm.fish#L33

qst0528:
Should I create its own feature branch for this update?
I'm not quite sure about git development conventions.

PR #637, opened 2022-10-24
Tidy up README and zsh shell config file

doolio:
Remove unnecessary semi-colon from vterm_prompt_end() function in zsh source code block.

Apply consistent naming to source code blocks (e.g. elisp vs emacs-lisp vs emacs etc.) to avoid confusion.  Remove superfluous whitespace before some source code block names.

Apply consistent whitespace either side of source code blocks improving readability when editing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ always want to use the vendored version as opposed to the one on you system, set
 
 Build the module with:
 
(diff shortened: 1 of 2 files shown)


Sbozzolo:
LGTM

Issue #567, opened 2022-01-06
Feature request: detect login shell over tramp


realcomplex:
Hi! Thanks for a great package; I'm using vterm a lot.

After a little bit of setup I also got vterm to mostly play nice with tramp, and now I can:
1. Use tramp to navigate to some remote server over ssh.
2. Open a vterm which will automatically open a shell on that server.

This is great, but there is one detail which is a little annoying for me: all tramp connections using the same method will result in the same shell being opened. This is configured with the `vterm-tramp-shells` variable; if it for example contains the entry `("ssh" "/usr/bin/zsh")`, then vterm will open a zsh shell on every server I connect to over ssh.

The problem of course is that I don't use the same shell on every computer I ssh to. Sometimes it's zsh, sometimes it's bash and sometimes something else. The most defensive option is to use something like "/bin/sh", but that's not very user-friendly. On the other hand, if vterm tries to open a shell which doesn't exist (say, `vterm-tramp-shells` contains `("ssh" "/usr/bin/zsh")` but zsh is not installed on a particular server), then vterm fails silently and doesn't open a terminal. The same happens when `vterm-tramp-shells` doesn't contain an entry for the current connection method and the default vterm shell doesn't exist on the remote.

I think it would be great if:
- the default login shell (i.e. the one specified in `/etc/passwd`) was opened automatically, or
- we could configure a list of shells, and vterm tries to open each one in turn. (For me, the list would look something like `("/usr/bin/zsh" "/usr/bin/bash" "/bin/sh")`.)

I can see that this would need a change to the `vterm--get-shell` function. I'm not proficient enough with tramp to implement the first option, but I could give the second option a try and submit a draft PR if you want. Of course, there are probably other solutions too.

Silex:
@realcomplex can you share what config you did to make it work smoothly with TRAMP?

Out of the box "it works" but things like up arrow doens't repeat history but sends `^[[A`

jvillasante:
Is this in the works?
Is there a workaround?

aagon:
@realcomplex 

> After a little bit of setup I also got vterm to mostly play nice with tramp, and now I can:
> 1. Use tramp to navigate to some remote server over ssh.
> 2. Open a vterm which will automatically open a shell on that server.

How did you do that ? Would you mind sharing what you did ?

Thanks in advance !

qhga:
> @realcomplex
> 
[...]

PR #638, opened 2022-10-29
Add a fallback shell command

rittelle:
This commit introduces a new `defcustom` that configures a fallback shell executable which is used if the selected shell (e.g. `vterm-shell`) is not executable.

The fallback is especially useful if the `vterm` is opened on a remote host over ssh/tramp which does not have the configured `vterm-shell` installed, but the user does not want to configure a different shell for all ssh hosts, as some may have the shell in `vterm-shell` installed.

This is my suggestion in response to running into #567.
I have only tested my solution on my local machine (by configuring `vterm-shell` to something invalid).

This change introduces breaking behavior as adding command line arguments to `vterm-shell` would have worked previously.  If this is something that should be supported by vterm, my suggestion would be to add support for arguments by allowing the user to configure `vterm-shell` to a list or to introduce an additional `defcustom`.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -180,6 +180,12 @@ the executable."
   :type '(alist :key-type string :value-type string)
   :group 'vterm)
 
+(defcustom vterm-fallback-shell "/bin/bash"
+  "The shell that gets run in the vterm if the selected shell is not available."
+  :type 'string
+  :group 'vterm)
+
+
 (defcustom vterm-buffer-name "*vterm*"
   "The basename used for vterm buffers.
 This is the default name used when running `vterm' or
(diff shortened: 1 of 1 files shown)


lexa:
What if instead of a customized fallback shell you could use the same shell as used by TRAMP. If tramp works on the remote, then you could be sure that the shell is also working. Geting the TRAMP shell is just a one-liner: 
```
            (with-connection-local-variables shell-file-name)
```
see:
https://github.com/akermu/emacs-libvterm/pull/652/commits/fcc4ec50c5493181397980216b5c3c697a60b8bf

PR #657, opened 2023-02-10
Don't allow moving cursor below the last prompt via mouse

FrostyX:
Without this patch, I am able to click somewhere below the last prompt line and move the cursor there. I don't think this was an intended feature, it feels more like a weird quirk.

We don't have any function telling us the point or line number of the last prompt, and it doesn't seem to be easy to implement, so I needed to workaround this.


FrostyX:
This is how I can currently move my cursor below the last prompt line by a mouse click.

vterm-cursor-below-last-prompt

This patch moves the cursor to the prompt line instead.

Issue #580, opened 2022-03-14
How to enable an audible bell?


jguzman-tech:
I'd like to get vterm to register audible beeps such was when tab-complete fails or when running `echo -e '\a'`. I don't think it has anything to do with my `.emacs` because I tried running with `--no-init` and then only loading the vterm package and I still wasn't hearing the bell.

Is it possible to configure this?

jguzman-tech:
Unrelated to the vterm package I saw that there's a way to set a visual bell in tmux.

`set-window-option -g visual-bell on`

It puts a notification in the tmux status line. Works for me when running commands in tmux within a vterm buffer.

pataquets:
I guess that either bell is completely ignored by libvterm or it's not being forwarded to Emacs.
Anyway, I would like to add that having a hook if this is ever passed to Emacs would be also very useful.

dwymark-celestron:
I've tested https://github.com/akermu/emacs-libvterm/pull/681 and it works as expected. Thanks to @jguzman-tech for contributing it.

Since it seems that PR #681 is not going to be merged any time soon, here are instructions you can use if you want to use this feature by patching it in.

## Patching instructions

### 1. Clone the Repository

```bash
# Clone emacs-libvterm to a convenient location.
mkdir -p ~/.emacs.d/lib
cd ~/.emacs.d/lib
git clone https://github.com/akermu/emacs-libvterm.git
cd emacs-libvterm
```

The following steps assume your working directory is `emacs-libvterm`.

### 2. Apply the Bell Functionality Patch

```bash
# Download the patch
curl -o bell-patch.diff https://patch-diff.githubusercontent.com/raw/akermu/emacs-libvterm/pull/681.diff

# Apply the patch
git apply bell-patch.diff
```

### 3. Compile the Library

```bash
mkdir -p build
cd build
cmake ..
make

# Verify the module was built
ls -la ../vterm-module.so
```

### 4. Configure Emacs

Add this to your Emacs configuration (e.g., `init.el`):

```elisp
(use-package vterm
  ;; Add custom vterm library to load-path
  :load-path "~/.emacs.d/lib/emacs-libvterm"
  :config
  ;; Optional: Custom bell handler for testing
  (defun my-vterm-bell-handler ()
    "Display 'Bell!' message when vterm bell is triggered."
    (message "Bell!"))
  
  ;; Override the default ding function in vterm-mode
  (defun my-vterm-mode-setup ()
[...]

PR #678, opened 2023-06-10
Add handling for the terminal bell

jguzman-tech:
This PR sets the `bell` callback from the `VTermScreenCallbacks` struct to a function that indirectly calls the `ding` function in emacs. This will play an audible bell by default.

A user could disable the ding by putting this in their init file:
```
; disable dings (in vterm mode only)
(defun disable-vterm-bell ()
  (setq-local ring-bell-function (lambda () nil)))
(add-hook 'vterm-mode-hook 'disable-vterm-bell)
```

This [documentation page](https://www.gnu.org/software/emacs/manual/html_node/elisp/Beeping.html) describes the behavior and user-options for `ding`.

Information is lost in some terminal applications when there is no terminal bell, for example scrolling to the bottom of a file in `less`, or reaching the top of your bash history when using Ctrl+R to cycle through history. I had queried about the terminal bell in #580. This code sets the previously unset `bell` callback to enable that feature.

### DIFF
--- a/elisp.c
+++ b/elisp.c
@@ -33,6 +33,7 @@ emacs_value Flist;
 emacs_value Fnth;
 emacs_value Ferase_buffer;
 emacs_value Finsert;
+emacs_value Fding;
 emacs_value Fgoto_char;
 emacs_value Fforward_char;
 emacs_value Fforward_line;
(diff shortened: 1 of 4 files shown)


ackerleytng:
Looking forward to this feature!

jguzman-tech:
will make a new PR with minor changes

PR #685, opened 2023-09-21
Allow different foreground/background colors for same ANSI index

iostapyshyn:
So far, vterm has been using `:foreground` attribute of its faces for regular colors (0-7), and `:background` attribute for bright colors (8-15). This is counter-intuitive and limits customization possibilities: In particular, it has been impossible to specify different color values, depending on whether the ANSI index is used for background or foreground.

Emacs' `ansi-color-*` and `term-color-*` faces allow specifying colors in this way by setting the `:foreground` and `:background` attributes respectively, but have been missing the bright half of the palette. The palette has been completed in Emacs 28 with new faces `ansi-color-bright-*` and `term-color-bright-*`.

This PR extends the vterm color handling with new bright color faces similarly to `term.el` and `ansi-color.el` and adjusts the `vterm--get-color` logic to use `:foreground` or `:background` depending on whether the index is used as background or foreground. It also simplifies handling of two special cases (default foreground with an underline and default background with inverse video attribute) mapped as special indices -11 and -12. Instead, the relevant attributes (underline and inverse-video) are passed to `vterm--get-color`, where these cases are handled in Lisp instead of C.

I have not found a backwards-compatible way of adding these new faces, hence the 28.1 bump (Emacs 25 is 7 years old anyway). Let me know if I am missing something and there's a way to stay backwards-compatible and whether it is desired after all.

### DIFF
--- a/elisp.c
+++ b/elisp.c
@@ -42,6 +42,7 @@ emacs_value Frecenter;
 emacs_value Fset_window_point;
 emacs_value Fwindow_body_height;
 emacs_value Fpoint;
+emacs_value Fapply;
 
 emacs_value Fput_text_property;
 emacs_value Fadd_text_properties;
@@ -182,9 +183,9 @@ void set_cursor_blink(emacs_env *env, bool blink) {
                (emacs_value[]){env->make_integer(env, blink)});
 }
 
-emacs_value vterm_get_color(emacs_env *env, int index) {
+emacs_value vterm_get_color(emacs_env *env, int index, emacs_value args) {
   emacs_value idx = env->make_integer(env, index);
-  return env->funcall(env, Fvterm_get_color, 1, (emacs_value[]){idx});
+  return env->funcall(env, Fapply, 3, (emacs_value[]){ Fvterm_get_color, idx, args });
 }
 
 void set_title(emacs_env *env, emacs_value string) {
--- a/README.md
+++ b/README.md
@@ -529,7 +529,9 @@ is a global minor mode in Emacs, you can use `M-x blink-cursor-mode` to toggle.
 
 ## Colors
 
-Customize the following faces to your liking.
+Customize the following faces to your liking. It is possible to specify
+different colors for foreground and background via the `:foreground` and
+`:background` attributes.
 
 - vterm-color-black
 - vterm-color-red
--- a/README.md
+++ b/README.md
@@ -529,9 +529,7 @@ is a global minor mode in Emacs, you can use `M-x blink-cursor-mode` to toggle.
 
 ## Colors
 
-Set the `:foreground` and `:background` attributes of the following faces to a
-color you like. The `:foreground` is ansi color 0-7, the `:background` attribute
-is ansi color 8-15.
+Customize the following faces to your liking.
 
 - vterm-color-black
 - vterm-color-red
@@ -541,6 +539,14 @@ is ansi color 8-15.
 - vterm-color-magenta
 - vterm-color-cyan
 - vterm-color-white
+- vterm-color-bright-black
+- vterm-color-bright-red
+- vterm-color-bright-green
+- vterm-color-bright-yellow
+- vterm-color-bright-blue
+- vterm-color-bright-magenta
+- vterm-color-bright-cyan
+- vterm-color-bright-white
 
 ## Directory tracking and Prompt tracking
 
--- a/elisp.h
+++ b/elisp.h
@@ -29,6 +29,7 @@ extern emacs_value Qrear_nonsticky;
 extern emacs_value Qvterm_prompt;
 
 // Emacs functions
+extern emacs_value Fapply;
 extern emacs_value Fblink_cursor_mode;
 extern emacs_value Fsymbol_value;
 extern emacs_value Flength;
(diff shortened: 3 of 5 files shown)


Sbozzolo:
Thank you very much for your pull request!

While Emacs 25 is 7 years old, Emacs 28.1 is a little more than 1 year old, so there are major distributions without it (Debian, RHEL, Ubuntu LTS).

I don't think we need to be backward compatible, but it would be good if we could stay compatible with older versions of Emacs. I haven't looked at the code, but I wonder if it possible to simply ignore the bright faces for Emacs<28.1? 

What do you think?

iostapyshyn:
The problem is here that I need to specify faces to inherit from for the new bright faces. Since the `term-color-bright-*` faces are missing prior to 28.1, this results in a following picture in old emacs versions:

image

Faces do allow specifying a list for the `:inherit` attribute ([ELisp Manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Face-Attributes.html))
> If a list of faces is used, attributes from faces earlier in the list override those from later faces.

Using this, I expected the face merging to fallback to the later face in the list if the earlier one is not present. This doesn't seem to be the case: Setting `:inherit` of `vterm-color-bright-{color}` faces to `(term-color-bright-{color} term-color-{color})` results in the same picture as above in Emacs 25.

One possible workaround would be inheriting from regular `term-color-{color}` faces for the new bright faces. This matches the unthemed behavior prior to the change (bright and regular colors being the same). The downside is that Emacs 28+ users would not profit from the new faces unless they customize vterm faces directly (or use a theme that does so for them).

protesilaos:
> @iostapyshyn: Faces do allow specifying a list for the `:inherit` attribute

Maybe something like this will suffice?

```elisp
(defface vterm-color-bright-red
  `((t :inherit ,(if (facep 'term-color-bright-red)
                     'term-color-bright-red
                   'term-color-red)))
  "Face used to render red color code.
The foreground color is used as ANSI color 1 and the background
color is used as ANSI color 9."
  :group 'vterm)

;; OR

(defface vterm-color-bright-red
  `((t :inherit ,(if (> emacs-major-version 28)
                     'term-color-bright-red
                   'term-color-red)))
  "Face used to render red color code.
The foreground color is used as ANSI color 1 and the background
color is used as ANSI color 9."
  :group 'vterm)
```

iostapyshyn:
> @protesilaos: Maybe something like this will suffice?

That's it, thanks! For some reason, the possibility to eval code in face specification has slipped my mind. I will shortly push the changes

elisp.c
```diff
@@ -182,9 +183,9 @@ void set_cursor_blink(emacs_env *env, bool blink) {
                (emacs_value[]){env->make_integer(env, blink)});
 }
 
-emacs_value vterm_get_color(emacs_env *env, int index) {
+emacs_value vterm_get_color(emacs_env *env, int index, emacs_value args) {
   emacs_value idx = env->make_integer(env, index);
-  return env->funcall(env, Fvterm_get_color, 1, (emacs_value[]){idx});
+  return env->funcall(env, Fapply, 3, (emacs_value[]){ Fvterm_get_color, idx, args });
```
Sbozzolo, line 188: Why do we need to `apply` (as opposed to funcalling the function?)

Sbozzolo:
Thank you! Can you also please share an example where the bright/non-bright colors are visually clearly distinct?

elisp.c
iostapyshyn: I added new properties in `vterm--get-color` as optional `&rest args` (IMHO it looks nicer that way, since these properties are optional after all).

Here, `apply` is used to map [a list constructed in `cell_rgb_color`](https://github.com/iostapyshyn/emacs-libvterm/blob/03e335ef462500ef71c26386720043febc201626/vterm-module.c#L857) for properties onto `&rest args` in `vterm--get-color`.

Thus, it's purely cosmetic. Alternatively, `vterm--get-color` could have a signature `(defun vterm--get-color (index args) ...)` instead, without the need to use `apply` in C:

```c
return env->funcall(env, Fvterm_get_color, 2, (emacs_value[]){idx, args});
```

iostapyshyn:
> @Sbozzolo: Can you also please share an example where the bright/non-bright colors are visually clearly distinct?

Here's an example:
2023-10-02T11:05:39,893483250+02:00

This is `modus-vivendi-tinted` with following overrides:

```lisp
(setq modus-themes-common-palette-overrides
      '((bg-term-black           "black")
        (fg-term-black           "gray20")
        (bg-term-black-bright    "gray40")
        (fg-term-black-bright    "gray60")

        (bg-term-red             bg-red-subtle)
        (fg-term-red             red)
        (bg-term-red-bright      bg-red-intense)
        (fg-term-red-bright      red-warmer)

        (bg-term-green           bg-green-subtle)
        (fg-term-green           green)
        (bg-term-green-bright    bg-green-intense)
        (fg-term-green-bright    green-cooler)

        (bg-term-yellow          bg-yellow-subtle)
        (fg-term-yellow          yellow)
        (bg-term-yellow-bright   bg-yellow-intense)
        (fg-term-yellow-bright   yellow-warmer)

        (bg-term-blue            bg-blue-subtle)
        (fg-term-blue            blue)
        (bg-term-blue-bright     bg-blue-intense)
        (fg-term-blue-bright     blue-warmer)

        (bg-term-magenta         bg-magenta-subtle)
        (fg-term-magenta         magenta)
        (bg-term-magenta-bright  bg-magenta-intense)
        (fg-term-magenta-bright  magenta-cooler)

        (bg-term-cyan            bg-cyan-subtle)
        (fg-term-cyan            cyan)
        (bg-term-cyan-bright     bg-cyan-intense)
        (fg-term-cyan-bright     cyan-cooler)

        (bg-term-white           "gray70")
        (fg-term-white           "gray80")
        (bg-term-white-bright    "gray90")
        (fg-term-white-bright    "white")))
```

Sbozzolo (approved):

Stebalien:
@Sbozzolo can we get this merged?

Sbozzolo:
Thanks for the ping!

PR #688, opened 2023-11-05
Enable reflow when supported

blahgeek:
term reflow is supported since libvterm v0.3. (the libvterm version in the cmakelists already includes this)

1. echo a long line of text
2. reduce the window width to be shorter than the line of the text
3. increase the window width to the original size

old behavior: the text is truncated and not recovered

new behavior: the text is reflowed


blahgeek:
Hmm there seems to be some crash issues. Still WIP

martinbaillie:
Would also _love_ this feature. I reckon many folks would, they just don't know it yet :-)

blahgeek:
Sorry folks, this PR is still a draft since I encountered several cases of random crashes while running with this patch and I haven't got the time to debug yet.

Thaodan:
Did you get any time to look into it?

Issue #692, opened 2023-12-09
vterm-module-compile could not detect build failure


sirikid:
`M-x version`:
```
GNU Emacs 30.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.37, cairo version 1.16.0)
```

In environment where CMake is present, but other dependencies are not, `vterm-module-compile` fails to detect a build failure and mistakenly reports build as successful.

This would not be much of a problem in interactive environment, but call to `vterm-module-compile` can also happen during Emacs initialization, in this case user will not see the ` *Install vterm* ` buffer and can have troubles switching to it because it is hidden.

<details>

<summary>*Messages* and *Install vterm* (picture)</summary>

![image](https://github.com/akermu/emacs-libvterm/assets/4840430/c6f2d2e7-4610-443f-8b98-8fbe9fd001b4)

</details>

PR #693, opened 2023-12-09
Move build directory management from Sh to Elisp

sirikid:
I moved creation of build directory and setting it as current from shell command to Elisp to get rid of masking CMake exit code by `cd`. Inspired by #692 but does not fully solve it.

```
(call-process "sh" nil nil nil "-c" "false; true") ;; 0
```

I also get rid of unconditional buffer popping, there is not much use in it without build failure.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -92,44 +92,30 @@ confirmation before compiling."
 (defvar vterm-install-buffer-name " *Install vterm* "
   "Name of the buffer used for compiling vterm-module.")
 
-(defun vterm-module--cmake-is-available ()
-  "Return t if cmake is available.
-CMake is needed to build vterm, here we check that we can find
-the executable."
-
-  (unless (executable-find "cmake")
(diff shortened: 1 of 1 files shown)


vterm.el
```diff
@@ -92,44 +92,30 @@ confirmation before compiling."
...
+           (file-name-directory (locate-library "vterm.el" t))))
+         (default-directory (expand-file-name "build" vterm-directory))
+         (make-commands (format "cmake -G 'Unix Makefiles' %s .. && make"
+                                vterm-module-cmake-args)))
+    (make-directory default-directory t)
+    (with-current-buffer (get-buffer-create vterm-install-buffer-name)
       (compilation-mode)
       (if (zerop (let ((inhibit-read-only t))
-                   (call-process "sh" nil buffer t "-c" make-commands)))
+                   (call-process "sh" nil t t "-c" make-commands)))
```
Thaodan, line 116: You could use create-process-shell-command here.

sirikid: Thanks, fixed

PR #694, opened 2024-01-02
update fish shell directory tracking function

nohzafk:
for current fish version `3.6.1`, the directory tracking instruction on README does not work.

set the `vterm_prompt_end` function to trigger whenever the current directory changes. This can be done using the **--on-variable PWD** hook.

This modification ensures that vterm_prompt_end is called every time the PWD variable changes, which occurs whenever user change directories.

With the above change, we don't need to call vterm_prompt_end from within fish_prompt anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,16 +588,8 @@ PS1=$PS1'\[$(vterm_prompt_end)\]'
 For `fish`, put this in your `~/.config/fish/config.fish`:
 
 ```fish
-function vterm_prompt_end;
+function vterm_prompt_end --on-variable PWD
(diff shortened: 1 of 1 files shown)


nohzafk:
check for running fish in vterm also works.

```fish
if [ "$INSIDE_EMACS" = vterm ]
    function vterm_prompt_end --on-variable PWD
        vterm_printf '51;A'(whoami)'@'(hostname)':'(pwd)
    end
end

```

Sbozzolo:
I am running fish 3.7 and the version in the README works, but I wouldn't mind updating to a cleaner version.

Is your proposed solution backward compatible?

Issue #655, opened 2023-02-04
Feature request: Open a "normal" shell over ssh


mpedramfar:
When I open a buffer over ssh with tramp and open vterm there, the shell that's opened is not exactly identical to the shell I would get if I open vterm locally and then ssh. In particular, the prompt is different and the shell history file is overriden with `tramp-histfile-override` (https://github.com/akermu/emacs-libvterm/issues/538). 
I'm not familiar with tramp, so I can't say for sure if these are the only things that have been changed.

Would it be possible to have the option of opening a "normal" shell instead? If this can be done easily by changing some variables, it would be great to have it in README so that users could see the set of all options that they can configure for remote shells.
Thanks

Bawaw:
You can set the remote shell by setting the the `vterm-tramp-shells` variable. 

E.g. if you want to use bash in your case: 
`(setq vterm-tramp-shells '(("ssh" "/bin/bash")))`

aramirezreyes:
I've been looking for this for a while, thanks! Where should I find this kind of documentation?

Thaodan:
> You can set the remote shell by setting the the `vterm-tramp-shells` variable.
> 
> E.g. if you want to use bash in your case: `(setq vterm-tramp-shells '(("ssh" "/bin/bash")))`

I think this works quite good however it would be better if the login shell from the target host would be used on the remote.

PR #706, opened 2024-03-06
Allow using the default login shell for remote connections

florommel:
For remote machines, let vterm start with the default login shell of the remote host.
This can be configured by specifying a the symbol 'login-shell as SHELL in vterm-tramp-shells. The current default behavior is not changed.
In addition, allow "*" as a wildcard method in vterm-tramp-shells.

This was requested in #655

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -172,7 +172,8 @@ the executable."
   :type 'string
   :group 'vterm)
 
-(defcustom vterm-tramp-shells '(("docker" "/bin/sh"))
+(defcustom vterm-tramp-shells
+  '(("ssh" login-shell) ("scp" login-shell) ("docker" "/bin/sh"))
   "The shell that gets run in the vterm for tramp.
 
 `vterm-tramp-shells' has to be a list of pairs of the format:
--- a/vterm.el
+++ b/vterm.el
@@ -176,7 +176,18 @@ the executable."
   "The shell that gets run in the vterm for tramp.
 
 `vterm-tramp-shells' has to be a list of pairs of the format:
-\(TRAMP-METHOD SHELL)"
+\(TRAMP-METHOD SHELL)
+
+Set SHELL to \\='login-shell to use the user's login shell on the host.
+The login-shell detection currently works for POSIX-compliant remote
+hosts that have the getent command (regular GNU/Linux distros, *BSDs,
+but not MacOS X unfortunately).
+
+You can specify an additional second SHELL command as a fallback
+that is used when the login-shell detection fails, e.g.,
+\\='((\"ssh\" login-shell \"/bin/bash\") ...)
+If no second SHELL command is specified with \\='login-shell, vterm will
+fall back to tramp's shell."
   :type '(alist :key-type string :value-type string)
   :group 'vterm)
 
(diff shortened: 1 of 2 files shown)


florommel:
I changed the wildcard from "*" to t, as it feels more idiomatic.

The last commit (c11336a) sets the default value of vterm-tramp-shells to `'((t login-shell))`. This should try to detect the login shell for all methods and fall back to Tramp's shell if this fails.

florommel:
This fixes #567

vterm.el
```diff
@@ -818,11 +828,34 @@ Exceptions are defined by `vterm-keymap-exceptions'."
...
+                          (with-current-buffer standard-output
+                            ;; Use a shell command here to get $LOGNAME.
+                            ;; Using the Tramp user does not always work
+                            ;; as it can be nil, e.g., with ssh host configs
+                            (process-file-shell-command
+                             "getent passwd $LOGNAME"
+                             nil (current-buffer) nil)))))
+               (shell (when shell
+                        (nth 6 (split-string shell ":" nil "[ \t\n\r]+")))))
+          (or shell second))
```
Sbozzolo, line 850: Thank you for your PR (and so sorry for the very very long turnaround time...)

What assumptions are being made here, and how robust is this with respect to different possible hosts (eg, different operating systems)?

florommel: Yes, I see, that's an important question.
I have updated `vterm--tramp-get-shell`. Now it should be pretty robust.

```elisp
    (if (or (eq first 'login-shell)
            (and (consp first)(eq (cadr first) 'login-shell)))
        (let* ((shell (ignore-errors
                        (with-output-to-string
                          (with-current-buffer standard-output
                            ;; Use a shell command here to get $LOGNAME.
                            ;; Using the Tramp user does not always work
                            ;; as it can be nil, e.g., with ssh host configs
                            (unless (= 0 (process-file-shell-command
                                          "getent passwd $LOGNAME"
                                          nil (current-buffer) nil))
                              (error "Unexpected return value"))
                            ;; If we have more than one line,
                            ;; the output is not the expected single passwd entry.
                            ;; Most likely, $LOGNAME is not set.
                            (when (> (count-lines (point-min) (point-max)) 1)
                              (error "Unexpected output"))))))
               (shell (when shell
                        (nth 6 (split-string shell ":" nil "[ \t\n\r]+")))))
          (or shell second))
      first)
```

- `$LOGNAME` should be defined in every POSIX-compliant system,
see https://pubs.opengroup.org/onlinepubs/007908775/xbd/envvar.html
If it is not defined, `vterm--tramp-get-shell` will return the fallback value or the default shell ("Unexpected output" -> `shell` is nil).
- `getent passwd` returns the passwd entry for the specified user, independently from the used name service (I also tested it with LDAP),
see https://en.wikipedia.org/wiki/Getent
The `getent` command is widely available in Unix systems [1].
However, MacOS X does not provide it. Thus for MacOS X hosts, `vterm--tramp-get-shell` will return the fallback or the default value ("Unexpected return value" -> `shell` is nil).

[1] from the FreeBSD manpage for getent:  "A getent command appeared in NetBSD 3.0, and was imported into FreeBSD 7.0. It was based on the command of the same name in Solaris and Linux."
https://man.freebsd.org/cgi/man.cgi?query=getent

So, it works for most Unix-like remote hosts, including GNU/Linux and the BSDs.
For MacOS X, the probing fails (since `getent` is missing), but the function should return the fallback value (not tested, since I have no Mac).
I don't know about Windows.. I guess with the Linux subsystem it should work.

If you deem it too unstable, I would argue for changing the default value of `vterm-tramp-shells` back to the original value but include the 'login-shell functionality as an optional feature.

Sbozzolo: Thank you. 

I think we can go ahead with this. Here are some comments to make the solution more robust (and cleanup a few things):
- Tramp supports all sorts of protocols and systems. While I expect ssh to be the most common (by far), we should not make too many assumptions about it. I suggest we turn the `login-shell` method you introduce to be the default for `ssh`, but let the user pick it for other protocols if they want to. 
- In this vain, I think it would be nice to check that we have a corresponding entry in `vterm-tramp-shells` for the method being used and error out when we don't 
- Restrictions and assumptions should be documented in the code (as you did in the previous message). More explanation is welcomed too (e.g., why `(nth 6`)? You could paste an example so that readers would konw what to expect)
- Please, format your code and check that docstrings are complete and accurate, including defining all the arguments function take
- Please, add an entry to the README.

Thank you again!

florommel: I have updated the PR.
Now, there is extensive in-code documentation and a extended docstrings. I also added a README entry with examples.

> I suggest we turn the login-shell method you introduce to be the default for `ssh`

I set it to default for `ssh` and `scp`, since `scp` are ssh servers in most cases. What do you think?

> In this vain, I think it would be nice to check that we have a corresponding entry in vterm-tramp-shells for the method being used and error out when we don't

I am not exactly sure what you mean by that.

Sbozzolo (approved):

nfuhler:
This feature would be quite useful to me and I just stumbled upon this pull request.
I was wondering about its current state since it seems to have been approved but not yet merged?

iostapyshyn:
I've been using @florommel's fork since March and haven't encountered any problems with it. Would like to have it merged as well.

PR #718, opened 2024-04-28
Support overriding SB_MAX at compile-time with -D defines

jahkeup:
This allows the user to select the scrollback buffer size via compiler preprocessor defines (instead of having to edit the source file) and falls back to the statically configured `SB_MAX` if there is no `-DSB_MAX=$VALUE` given.


Sbozzolo:
Thank you!

PR #722, opened 2024-06-18
Do our best to restore tab characters

Dieken:
```
$ echo -e 'abc\t\t\tabc'
  # use mouse to select and copy the output, it should contain three tabs,
  # not expand to spaces.

$ touch 2024中华人民共和国
$ for i in `seq 1 20`; do touch text-text-text-text-text-text-$i; done 
$ gls
$ gls -C | cat
  # It should align correctly.
```

### DIFF
--- a/vterm-module.c
+++ b/vterm-module.c
@@ -316,6 +316,7 @@ static void refresh_lines(Term *term, emacs_env *env, int start_row,
 
   for (i = start_row; i < end_row; i++) {
 
+    int consecutive_empty_cells = 0;
     int newline = 0;
     int isprompt = 0;
     for (j = 0; j < end_col; j++) {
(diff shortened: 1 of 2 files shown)


Sbozzolo:
Does this assume that any 8 consecutive spaces are always a tab?

Dieken:
> Does this assume that any 8 consecutive spaces are always a tab?

No,  consecutive **empty** cells reaching tab stop are converted to tab.

Empty cell means `cell.chars[0] == 0`,   not `cell.chars[0] == ' '`.

Dieken:
@Sbozzolo  @jixiuf  could you merge this?

Dieken:
@Sbozzolo @jixiuf could you merge this?

PR #739, opened 2024-11-12
Fixes #731: Prevent the display of truncation characters ($)

### LINKED ISSUE #731: Using the built-in text-scale-mode causes vterm to have display issues
When using Emacs vterm with ncurses programs that change the background color, a `$` symbol appears at the end of some lines when `(text-scale-increase)` and `(text-scale-decrease)` are used to change the font size.

Steps to Reproduce:
- Open Emacs and start vterm (M-x vterm).
- Run an ncurses-based program within vterm (vim) and make sure that the background background color of the ncurses program is different (for example the Vim command `:color blue` or `:color murphy`).
- Call the Emacs functions `(text-scale-increase)` and/or `(text-scale-decrease)` to change the font size.  
- Observe the display for any lines ending with a $ symbol or sometimes just with a different background on the right.

Expected Behavior:
The ncurses program should display correctly without any extraneous `$` symbols at the end of lines.

Actual Behavior:
The `$` symbol appears at the end of some lines when the ncurses program changes the background color. This behavior is not expected and disrupts the display.

Additional Information:
- Emacs version: 29.4
- vterm version: commit ref: d9ea29fb10aed20512bd95dc5b8c1a01684044b1
- Operating System: Tested on Gentoo, Arch Linux, and Debian (same issue)

jamescherti:
This pull request fixes the issue #731 where the $ symbol appears at the end of some lines.

It usually happens when the user calls `text-scale-increase` or `text-scale-decrease` to change the font size.

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -813,6 +813,14 @@ Exceptions are defined by `vterm-keymap-exceptions'."
   (vterm--set-pty-name vterm--term (process-tty-name vterm--process))
   (process-put vterm--process 'adjust-window-size-function
                #'vterm--window-adjust-process-window-size)
+
+  ;; Set the truncation slot for 'buffer-display-table' to the ASCII code for a
+  ;; space character (32) to make the vterm buffer display a space instead of
+  ;; the default truncation character ($) when a line is truncated.
+  (let* ((display-table (or buffer-display-table (make-display-table))))
+    (set-display-table-slot display-table 'truncation 32)
+    (setq buffer-display-table display-table))
+
   ;; Support to compilation-shell-minor-mode
   ;; Is this necessary? See vterm--compilation-setup
   (setq next-error-function 'vterm-next-error-function)


Sbozzolo:
Thank you! Can you please explain how this code works?

jamescherti:
Hello @Sbozzolo,

This line creates a local variable, `display-table`, that will hold the display table for the buffer:
```elisp
(let* ((display-table (or buffer-display-table (make-display-table)))):
```

This line modifies the display-table (local variable) by setting the slot for truncation to the value 32, which is the ASCII code for a space character:
```elisp
(set-display-table-slot display-table 'truncation 32):
```
(By setting the truncation slot to 32, it ensures that when a line is truncated, a space is shown instead of the usual truncation character `$`.)

This sets the buffer-local variable `buffer-display-table` to the modified display-table:
```elisp
(setq buffer-display-table display-table)
```

Sbozzolo (approved):

jamescherti:
You are very welcome, @Sbozzolo!

PR #753, opened 2025-01-11
README: link the issue of running on Windows

Hi-Angel:
It seems there is demand for making this work on Windows, so would be nice if the text that says it doesn't work would be linked to the issue that contains the research. This will allow interested people *(if such appear)* to avoid researching everything from scratch, but rather basing their work on the information already done by others.

This came up on SE: https://emacs.stackexchange.com/questions/82896/error-use-package-vterm-catch-searching-for-program-no-such-file-or-direct/82909#82909


Sbozzolo:
Thank you!

PR #760, opened 2025-02-25
Add vterm-ignore-cursor-change

Azkae:
Add an option to ignore application request to modify cursor shape or visibility.

I use [meow](https://github.com/meow-edit/meow), which uses cursor shape to differentiate between normal and insert mode, so I prefer if the application cannot change the cursor shape.


fredericgiquel:
This MR is useful for `evil` too, particularly with version 4.0 of `fish` shell which regularly changes the cursor shape.

The patch is small. I tested it and it seems to work well.
Is there anything preventing the merge ?

PR #770, opened 2025-06-15
Fixes #729: Ignore libvterm string with length -1

### LINKED ISSUE #729: vterm shell escape sequences crash emacs
Messing around with vterm escapes, I found that certain sequences cause emacs to crash. A simple one is `printf "\033];\033"`, but just about anything between the two escape characters has the same result.

The crash report shows exception type EXC_BAD_ACCESS (SIGABRT), with the access address pointing to a STACK GUARD memory region. This suggests vterm is trying to access memory outside a stack-allocated object.

GNU Emacs 29.3 (build 1, aarch64-apple-darwin21.6.0, NS appkit-2113.60 Version 12.6.6 (Build 21G646)) of 2024-03-24, installed from https://emacsformacos.com
vterm 20240705.1533, installed from melpa

strager:
libvterm sometimes gives us a string fragment with length 1073741823 [1]. When this happens, we attempt to read 1073741823 bytes, overrunning the buffer and causing a segfault.

Fix the crash by adding a check for -1. Because the number is not exactly -1 (because VTermStringFragment.len is a 30-bit bit field at least in my copy of libvterm), do some C trickery to get the -1 value for comparison.

[1] This -1 length might be a bug in libvterm. I did not investigate further.

### DIFF
--- a/vterm-module.c
+++ b/vterm-module.c
@@ -1149,6 +1149,12 @@ static VTermParserCallbacks parser_callbacks = {
 };
 #else
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wbitfield-constant-conversion"
+/* -1 is represented as ~0 in the bit field. */
+static const size_t invalid_string_fragment_len = (VTermStringFragment){.len = (size_t)-1}.len;
+#pragma clang diagnostic pop
+
 static int osc_callback(int cmd, VTermStringFragment frag, void *user) {
   /* osc_callback (OSC = Operating System Command) */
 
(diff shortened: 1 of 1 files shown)


Sbozzolo:
What happens when `frag.len == invalid_string_fragment_len`?

strager:
Are you asking what exactly causes frag.len to be invalid_string_fragment_len? I don't know, but the test case in #729 triggers it.

Are you asking what the code does when frag.len == invalid_string_fragment_len? It'll behave as if frag.len == 0. But that's not 100% true, because the branch on 1176 won't ever be true. Maybe the code would be clearer if I wrote `if (frag.len == invalid_string_fragment_len) frag.len = 0;` at the top of the function?

Sbozzolo:
> Are you asking what the code does when frag.len == invalid_string_fragment_len? It'll behave as if frag.len == 0. But that's not 100% true, because the branch on 1176 won't ever be true. Maybe the code would be clearer if I wrote if (frag.len == invalid_string_fragment_len) frag.len = 0; at the top of the function?

Yes, to me that would be clearer.

PR #777, opened 2025-09-27
Enable Windows support by using a conpty-proxy

xhcoding:
The conpty-proxy launches processes through conpty and redirects their input/output to its own stdin and stdout.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,16 +92,24 @@ endif()
 
 # Build conpty-proxy
 if(WIN32)
-  ExternalProject_add(conpty-proxy
-    GIT_REPOSITORY https://github.com/xhcoding/conpty-proxy.git
-    GIT_TAG v1.0.0
-    INSTALL_DIR ${CMAKE_CURRENT_SOURCE_DIR}
-    INSTALL_COMMAND
-        COMMAND ${CMAKE_COMMAND} -E copy
-        <BINARY_DIR>/conpty_proxy.exe
-        <INSTALL_DIR>/conpty_proxy.exe
-  )
-  add_dependencies(vterm-module conpty-proxy)
+
+  find_program(CONPTY_PROXY_PATH conpty_proxy.exe)
+
+  if (CONPTY_PROXY_PATH)
+    message(STATUS "Use system conpty: ${CONPTY_PROXY_PATH}")
+  else()
+    ExternalProject_add(conpty-proxy
+      GIT_REPOSITORY https://github.com/xhcoding/conpty-proxy.git
+      GIT_TAG v1.0.0
+      INSTALL_DIR ${CMAKE_CURRENT_SOURCE_DIR}
+      INSTALL_COMMAND
+      COMMAND ${CMAKE_COMMAND} -E copy
+      <BINARY_DIR>/conpty_proxy.exe
+      <INSTALL_DIR>/conpty_proxy.exe
+    )
+    add_dependencies(vterm-module conpty-proxy)
+  endif()
+
 endif()
 
 add_library(vterm STATIC IMPORTED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,23 +92,15 @@ endif()
 
 # Build conpty-proxy
 if(WIN32)
-
-  find_program(CONPTY_PROXY_PATH conpty_proxy.exe)
-
-  if (CONPTY_PROXY_PATH)
-    message(STATUS "Use system conpty: ${CONPTY_PROXY_PATH}")
-  else()
-    ExternalProject_add(conpty-proxy
-      GIT_REPOSITORY https://github.com/xhcoding/conpty-proxy.git
-      GIT_TAG v1.0.1
-      INSTALL_DIR ${CMAKE_CURRENT_SOURCE_DIR}
-      INSTALL_COMMAND
-      COMMAND ${CMAKE_COMMAND} -E copy
-      <BINARY_DIR>/conpty_proxy.exe
-      <INSTALL_DIR>/conpty_proxy.exe
-    )
-    add_dependencies(vterm-module conpty-proxy)
-  endif()
+  ExternalProject_add(conpty-proxy
+    GIT_REPOSITORY https://github.com/xhcoding/conpty-proxy.git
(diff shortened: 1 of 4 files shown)


Sbozzolo:
I don't have access to windows machines. Would anyone be interested in testing this?

Master-Hash:
> I don't have access to windows machines. Would anyone be interested in testing this?

I confirm it works, but documentation and config need polish.

e.g. The package relies on conpty_proxy.exe, whose path seems not configurable. 

图片

xhcoding:
> > I don't have access to windows machines. Would anyone be interested in testing this?
> 
> I confirm it works, but documentation and config need polish.
> 
> e.g. The package relies on conpty_proxy.exe, whose path seems not configurable (default: the same path as vterm.el, see https://github.com/xhcoding/emacs-libvterm/blob/master/vterm.el#L1465-L1467).
> 
> A working snapshot: 图片
 I've added a new variable to specify the path to conpty_proxy.exe

kiennq:
> Maybe try to install the pre-built version of libvterm by pacman

Somehow it doesn't work for native comp version of Emacs on Windows. The module loading kept failing.

kiennq:
@xhcoding, when I use interactive command line tool (ex: copilot-cli or python) inside `vterm`, sometimes it spawns a new conpty process and hangs Emacs. Do you know if that's fixable?

kiennq:
> @xhcoding, when I use interactive command line tool (ex: copilot-cli or python) inside `vterm`, sometimes it spawns a new conpty process and hangs Emacs. Do you know if that's fixable?

I found out the reason why. It's because vterm creates a new conpty process and wait for its result when doing resizing. Since the overhead is high, it makes the whole Emacs slow as long as the vterm is in display. Also, if the new conpty process is stuck the whole Emacs will hang without be able to quit using Ctrl + g (as I observed).

1. It's better to just call the external process without waiting for its result (notify) and use the passed `width, height` in Emacs immediately. Adding a debounce on creating process is also nice since it lowers the overhead of creating a new process in Windows as well.
2. Alternatively, we can use `call-process` and `accept-process-output` to wait for external process to finish, that will allow us at least to quit with Ctrl + g when Emacs stuck.

Here is a fix using approach 1 https://github.com/kiennq/emacs-libvterm/commit/0e0e179cdd1ac8d46e2f1657464e10b0afbe7185

lehaworker:
> > @xhcoding, when I use interactive command line tool (ex: copilot-cli or python) inside `vterm`, sometimes it spawns a new conpty process and hangs Emacs. Do you know if that's fixable?
> 
> I found out the reason why. It's because vterm creates a new conpty process and wait for its result when doing resizing. Since the overhead is high, it makes the whole Emacs slow as long as the vterm is in display. Also, if the new conpty process is stuck the whole Emacs will hang without be able to quit using Ctrl + g (as I observed).
> 
> 1. It's better to just call the external process without waiting for its result (notify) and use the passed `width, height` in Emacs immediately. Adding a debounce on creating process is also nice since it lowers the overhead of creating a new process in Windows as well.
> 2. Alternatively, we can use `call-process` and `accept-process-output` to wait for external process to finish, that will allow us at least to quit with Ctrl + g when Emacs stuck.
> 
> Here is a fix using approach 1 [kiennq@0e0e179](https://github.com/kiennq/emacs-libvterm/commit/0e0e179cdd1ac8d46e2f1657464e10b0afbe7185)

Your version works! Laggy, but usable.

leuven65:
> Here is a fix using approach 1 [kiennq@0e0e179](https://github.com/kiennq/emacs-libvterm/commit/0e0e179cdd1ac8d46e2f1657464e10b0afbe7185)

Nice fixing!

One remark: 
Maybe it is better to define "vterm--conpty-proxy-resize-timer" as buffer local variable, as same as "vterm--process", as each vterm session needs to have its own timer.

Some small adapt:
https://github.com/leuven65/emacs-libvterm/commit/aa1dcb52b17415c21d07811e672e721623fefdc7
https://github.com/leuven65/emacs-libvterm/commit/a6ac8ea15f932ce6e2b13e14266db3095a464b09

kiennq:
> Maybe it is better to define "vterm--conpty-proxy-resize-timer" as buffer local variable, as same as "vterm--process", as each vterm session needs to have its own timer.

Nice catch.

heheda123123:
Three weeks have passed. Is there any progress?

maksim-brylev:
> Three weeks have passed. Is there any progress?

You can use it right now, today, and then upstreaming shouldn't bother you that much.

mateialexandru:
Why is https://github.com/xhcoding/conpty-proxy needed? why can't we just have Windows APIs / expand on cmdproxy that Emacs ships?

kiennq:
`cmdproxy` is a shell which wrap the current `cmd` to allow passing command via `-c`. `conpty-proxy` is a shim layer that translate Windows shell output -> VT sequences that are compatible with `libvterm`. I think they should be separated, we can use any shell with vterm, not only `cmdproxy`.

Btw, I've move `conpty-proxy` directly to `vterm-module` here https://github.com/kiennq/emacs-libvterm
That would eliminate middle process and probably a bit smoother

maksimbrylev2004:
This breaks TRAMP on Linux remote boxes

pkradiator:
@kiennq I tried building https://github.com/kiennq/emacs-libvterm but when I do `cmake -S . -Bbuild` I get
```
CMake Error at CMakeLists.txt:163 (set_target_properties):
set_target_properties called with incorrect number of arguments.
```
The line 163 is 
```
set_target_properties(termiwin PROPERTIES IMPORTED_LOCATION ${LIBTERMIWIN_LIBRARY})
```

kiennq:
@pkradiator Can you check the build workflow in that repo? It's used to build the release binaries

pkradiator:
@kiennq Can you point out specifically what you are referring to? (I am a Noob). I followed all the instruction here [vterm and Windows](https://github.com/kiennq/emacs-libvterm?tab=readme-ov-file#vterm-and-windows)

kiennq:
Here https://github.com/kiennq/emacs-libvterm/blob/master/.github/workflows/build_module.yml, line 118-119

PR #786, opened 2025-10-25
Set up minimal GH Action to build the code

mgeisler:
This adds a very small smoke test to help ensure that GitHub PRs build and that the module loads.

Emacs is installed via Nix, despite the build machine being an Ubuntu VM. This seemed like the easiest and fastest way to get different versions of Emacs running on the VM: it only takes a few seconds to pull down the pre-compiled binaries.

The build uses the version of `libvterm` that CMake pulls down.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,48 @@
+name: Build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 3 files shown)


mgeisler:
By the way, the Nix part can probably be taken out if it's okay to just test using the version of Emacs found on the GH runners.

PR #788, opened 2025-11-19
Fix return type of `string_bytes`

mgeisler:
The function actually returns `ptrdiff_t` (and the callers already expected that type).


Sbozzolo:
Thank you!

PR #790, opened 2025-11-29
make the startup shell a variable to fix macos issues w/ posix

adonh:
* for macos, /bin/sh is dash by default, so allow user customization to be /bin/bash instead

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -195,6 +195,11 @@ fall back to tramp's shell."
   :type '(alist :key-type string :value-type string)
   :group 'vterm)
 
+(defcustom vterm-start-shell "/bin/sh"
+  "The shell that starts the vterm shell."
+  :type 'string
+  :group 'vterm)
+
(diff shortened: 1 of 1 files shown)


Sbozzolo:
What problem were you running into with `/bin/sh`?

adonh:
> What problem were you running into with `/bin/sh`?

apologies- didn't notice that my notifications were not working.

unfortunately, I don't quite recall the specific reason I had issues- it was something like vterm launching a shell script, e.g. claude-code, where that script was expecting bash-isms to work, but because the start shell was POSIX (or b/c it's `dash` on macos), it got unhappy.

I'm actually tinkering around this some more- to allow a `-l` shell to be the launching shell (to follow the stupid macos behavior w/ shells). As it's written, because we're just passing a single string to `make-process`, it's not possible to inject additional command-line arguments to the start shell in front of the `-c'. I'll make a separate PR for that version and you all can think about accepting either or not.

TIA!

PR #795, opened 2026-02-27
initialize `vterm-shell` the same way as `shell-mode` does

adonh:
- also have a `vterm-start-shell` to customize starting the `vterm-shell`
- optionally both can default to how `shell-mode` defaults

### DIFF
--- a/vterm.el
+++ b/vterm.el
@@ -167,9 +167,12 @@ the executable."
 
 ;;; Options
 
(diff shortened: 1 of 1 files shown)


adonh:
perhaps you may kindly review this fuller version over https://github.com/akermu/emacs-libvterm/pull/790 ?
TIA!

PR #798, opened 2026-04-05
feat: enable libvterm reflow on window width resize

### LINKED ISSUE #179: re-wrap line when window resized

milanglacier:
Fixes #179 

# Problem

```text
Before resize:
  $ some_command
  error: something went wrong while processing /very/long/path/to/a/file/that/keeps/going

After shrinking the Emacs window:
  $ some_command
  error: something went wrong while processing /very/long/path/to/
  a/file/that/keeps/going

In some cases the line is not only wrapped, but part of the visible output is clipped,
so the buffer no longer shows the original full line cleanly after resizing.
```

# Solution

This is similar to the PR #688, but this PR attempts to address the problem that the original PR is mentioning:

> Sorry folks, this PR is still a draft since I encountered several cases of random crashes while running with this patch and I > haven't got the time to debug yet.

With reflow enabled, `sb_pushline` can fire on width-only resizes, not just when the window gets shorter. The old code in `vterm-module.c` assumed any scrollback push during resize meant the height had decreased, so it always trimmed `term->lines_len`. Do that enough times and the module's line metadata falls out of sync with libvterm—eventually segfaulting in `vterm--window-adjust-process-window-size`.

Now we only shrink the metadata when we're actually losing rows from a height decrease. Width-only reflow pushes leave the line count alone.


milanglacier:
Example video with reflow:


https://github.com/user-attachments/assets/39d43873-27c7-4236-8f34-72c7a66696d1

Sbozzolo:
Thank you very much!

I tested the PR with a clean emacs and vterm and I think there is still some hiccup with the point and with lines being repeated, see screencast.

https://github.com/user-attachments/assets/86ed8e8a-69f1-411a-a349-3dc946ac1388

milanglacier:
Thank you for the update!

I think the repeated lines problem you mentioned during window width resize, are coming from older problem, I have encountered similar issue in older version of vterm before the reflow is implemented:

see the following screenshot (where I am using an older version of vterm):

With that being said, I can definitely look into that issue, and try to fix it at a new PR. But I think the problem is coming from the older codebase, not a issue with this PR that enables the reflow.

https://github.com/user-attachments/assets/267a7b62-6ef7-49f2-ba79-97bcf772bb44

milanglacier:
Also notice that Neovim exhibits a similar issue with repeated lines.  Given that Neovim displays the same behavior, I thinkthis may be an upstream vterm issue?

https://github.com/user-attachments/assets/07ea14d0-228d-43aa-b4fa-c0e8c63d60fe

dickmao:
Bros, it's a bug in screen.c, fixed in my fork.  This is why you have to vendor the libvterm source code.
reflow

milanglacier:
Any updates on this? While this patch isn’t perfect due to the upstream bug, it still provides a desirable improvement over the current behavior. At present, text extending beyond the screen width is permanently truncated rather than reflowed. Merging this feature would therefore benefit most users.

Sbozzolo:
> Any updates on this? While this patch isn’t perfect due to the upstream bug, it still provides a desirable improvement over the current behavior. At present, text extending beyond the screen width is permanently truncated rather than reflowed. Merging this feature would therefore benefit most users.

Yes, I agree.

vterm-module.c
Sbozzolo, line 118: Is there a mirror fix to be applied here?

I am thinking that with reflow enabled `term_sb_pop` can fire during width-only increases, so that each pop grows `lines_len` past `height`, and `term_resize` won't correct it since `rows == height` for a width-only resize.

milanglacier:
pushed a commit on the sb_pop event.

vterm-module.c
```diff
@@ -69,8 +72,9 @@ static int term_sb_push(int cols, const VTermScreenCell *cells, void *data) {
   sbrow->info = term->lines[0];
   memmove(term->lines, term->lines + 1,
           sizeof(term->lines[0]) * (term->lines_len - 1));
-  if (term->resizing) {
```
Sbozzolo, line 72: This variable is now unused, so we could remove it.

Sbozzolo:
Thank you, looks good to me. Only one tiny comment left.

milanglacier:
ok. Pushed a new commit removed the resizing flag.

Sbozzolo:
Thank you for pushing this through!

juhp:
Thanks

The biggest problems with wrapping I face are:
1. urls getting split across lines: I am using `ffap` (`find-file-at-point`) - is there a better package now?
  - basically I end up marking and copying long urls by hand as my workaround
2. narrowing the width of vterm causes the end of lines to get lost when widening again
  - this is improved with this change, but the new wrapping causes the top displayed lines (that flow out of the window) still to lose their ends
3. now I see some artifact characters when widening dnf update output (dunno if it is related with ansi codes getting cut or something else).

I think ghostel can handle these iirc, but it has some other issues (at least for my current config)

*(perhaps I should have commented in the issue instead)*

